PR #30, opened 2015-07-23
Copy or not   performance comparison

KjellKod:
Making pull request for visibility, then closing it again. 

10 sinks: all writing 2 million log entris to disk. Same "application" overhead for LOG calls,. approx 4us. 
However, this version was about 20% slower in background processing compared to the original one which keeps it on the heap. Cache issues?

I think it's time to use vtune.

### DIFF
--- a/src/filesink.cpp
+++ b/src/filesink.cpp
@@ -49,9 +49,9 @@ namespace g3 {
    }
 
    // The actual log receiving function
-   void FileSink::fileWrite(LogMessageMover message) {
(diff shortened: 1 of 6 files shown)


KjellKod:
:-1:   Just for fun to see if copying was faster than re-creating on the heap.   It wasn't. 

The queue overhead is what overshadows everything else. Time to test a different queue on the Actor -> Actor message passing which is ALWAYS single thread -> single thread. (only the first Actor receives on a Multiple Producer)

PR #47, opened 2015-09-06
add option ADD_BUILD_SHARED

ctapmex:
Now for windows build shared dll not working. this option disable it.
also  - if g3log used as a git submodule, then in visual studio project g3log_shared may not be necessary


KjellKod:
Nice.

One question: if you are on Windows, should the default be to have *not* shared? 
With this pull request the default is shared, for all platforms.

I think it is fine the way it is, just throwing it out there

ctapmex:
I agree. change the logic

KjellKod:
:+1:

PR #51, opened 2015-09-16
For Craig

KjellKod:


KjellKod:
Craig will do a pull request with the fixes also for the two failing tests

PR #52, opened 2015-09-16
Adding the function name and saving the day

craig-cogdill:


test_unit/test_io.cpp
```diff
@@ -541,7 +541,7 @@ TEST(CustomLogLevels, AddFatal) {
 
    std::string file_content = readFileToText(logger.logFile());
    std::string expected;
-   expected += "DEADLY [test_io.cpp:" + std::to_string(line);
+   expected += "DEADLY [test_io.cpp->" + std::string(__FUNCTION__) + ":" + std::to_string(line);
```
KjellKod: This is fine. U shouldn't need the std::string wrapping though. 
Did it not work without it?

KjellKod:
:+1:

test_unit/test_io.cpp
craig-cogdill: No. Invalid operands to + argument

PR #63, opened 2015-11-24
fix msvc2013 compilation issue of internal::g_log_level_status on res…

cstamatopoulos:
…et()


KjellKod:
:+1: the pull request failed the build but I fixed that in the prior pull request

Issue #13, opened 2015-06-17
Missing import lib on windows dll


pendenaor:
Hi,

When compiling the VC2013 solution, there's no g3logger_shared.lib generated.

Looking into the solution, i see no definition file or /EXPORT compiler flag or __declspec(export) into g2log.hpp, so VC2013 does not generate import lib (see https://msdn.microsoft.com/en-us/library/67wc07b9.aspx for more information).

KjellKod:
Thanks 
I will look into it. According to the Cmake file it should be generated so something is amiss https://github.com/KjellKod/g3log/blob/master/Build.cmake

pendenaor:
I'm a cmake noob, but maybe you could look at http://www.cmake.org/Wiki/BuildingWinDLL

KjellKod:
@pendenaor, what CMake arguments did you used to generate the solutions file?

KjellKod:
I have verified with cmake and the latest on g3log that the g3logger_shared.dll is generated

To reproduce I followed the instructions in the g3log/CMakeLists.txt
(in g3log)
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release -G "Visual Studio 12" ..
This will generate the solutions file g3log.sln

To build from command line:
msbuild g3log.sln /p:Configuration=Release

which generates 
in the Release directory
g3logger_shared.dll among other files.

pendenaor:
"g3logger_shared.dll" is generated, no problem with that, but not his import library "g3logger_shared.lib" which export functions or classes. This behavior is specific to the windows platform and vc compiler (gcc make it public by default).

KjellKod:
Aha. Thanks for the clarification.

KjellKod:
I agree that this should be fixed. I'm off on vacation for close to two weeks so I likely won't have time to look at this until after.

Pull requests are always welcomed ;)

KjellKod:
I have thought long and hard about this one. I'm OK with this being implemented but I will not do it myself. This one I will leave for someone on the Windows side to implement and do a pull request for it.

I'm less on the Windows side than on OSX and Linux side and a feature like this is better made by someone who has a good working knowledge of the ins and outs of DLLs.

ctapmex:
to realize this in the beginning you need to alter the structure in the same classes. 
LogWorkerImpl, LogCapture and some other

jkhoogland:
there is some logic coming up in cmake 3.4, 
[...]

PR #65, opened 2015-11-27
#13 generate import lib in msvc using cmake >= 3.4

cstamatopoulos:


KjellKod:
Looks good. Can I try this on msvc 2013?

I have one request. Can you supply an example with this pull request that takes advantage of the exported g3log library? Kind of a proof-of-concept for the users that need this?

cstamatopoulos:
Yes you can try it with msvc2013. You just need cmake 3.4 of course

I can provide an example but the whole point really is the the import lib.
Once you have that you just use it identically with your static lib but you just need the dll in the same folder with the .exe.
So not really sure what it adds.

KjellKod:
I finally got hand on a windows computer to test this with. 

I tried building the fatal examples (example/Example.cmake) and changed the linking for the examples from ```target_link_libraries(g3log-FATAL-choice ${G3LOG_LIBRARY})``` to  ```target_link_libraries(g3log-FATAL-choice ${G3LOG_SHARED_LIBRARY})``` 

```cmake  "Visual Studio 12" -DADD_BUILD_WIN_SHARED=ON ..```

It seems to work great! Thanks man!
:+1:

Unfortunately I noticed that some unit tests are broken(not this pull request's fault)  so before I merge this I will make sure that the unit tests are up and running on Windows as well.

KjellKod:
This pull request will also close issue:  https://github.com/KjellKod/g3log/issues/13

I will follow up with another pull request to fix unit tests for Windows and to deal with linking the examples with the dll if the right build option is given

PR #78, opened 2016-02-15
add "_d" postfix for msvc debug builds

cstamatopoulos:


Build.cmake
```diff
@@ -85,6 +85,10 @@ ENDIF()
    ELSE()
          list(REMOVE_ITEM SRC_FILES  ${LOG_SRC}/crashhandler_windows.cpp ${LOG_SRC}/g3log/stacktrace_windows.hpp ${LOG_SRC}/stacktrace_windows.cpp)
    ENDIF (MSVC OR MINGW)
+   
+   IF (MSVC) 
+         SET(CMAKE_DEBUG_POSTFIX "_d")
```
KjellKod: Great I will try it out.

KjellKod: I was hesitant to add this but it seems that the overwriting of library files (debug/release) is a common issue for Windows. 

I've been away (injured) from coding from some time so I'm just getting back to this now.  I still have to verify that this works in conjunction with g3sinks PR https://github.com/KjellKod/g3sinks/pull/20

KjellKod: for fun I tried to set this for Linux (commenting out the MSVC check). It failed. 

This explains how it should be set but I have not yet tried it. I will try to see if this (see link) actually does anything or if the link below has to be adhered to: 

http://stackoverflow.com/questions/28893450/how-to-get-debug-postfix-in-executable-name

KjellKod:
I have lost access to Windows dev environment. Unfortunately this means that this pull request has to be pending unless someone else will verify it.

PR #76, opened 2016-01-26
Allow user to set the logger identifier -- default = ".g3log."

jkhoogland:

PR #82, opened 2016-02-17
Repace g3log in path

KjellKod:
potential replacement for https://github.com/KjellKod/g3log/pull/76 so that @jkhoogland  does not have to update it.


KjellKod:
Thanks to  @jkhoogland  for adding this:  https://github.com/KjellKod/g3log/pull/76

PR #90, opened 2016-05-08
Added TimestampProvider API Enabling Milli / Micro Second Support in Timestamps

gimesketvirtadieni:
This is a pull request for a functionality for dynamic sink removal


KjellKod:
Thanks for providing this.  I haven't had time to look at in until now but am promptly reviewing it. 
thanks

src/g3log/logmessage.hpp
```diff
@@ -87,7 +88,8 @@ namespace g3 {
       // Complete access to the raw data in case the helper functions above
       // are not enough.
       //
-      std::time_t _timestamp;
+      // TODO: unique_ptr should be considered instead of shared_ptr
+      std::shared_ptr<Timestamp> _timestamp;
```
KjellKod: I do not see the reason for having a unique or shared pointer for this. By introducing a shared_ptr you will also introduce quite a bit of overhead and possibility of cache misses.   

Since the pull request targets time sensitive applications I do not think this is in your interest. _(it's not in mine)_

KjellKod:
Hi @gimesketvirtadieni, I see the what you did here and I think it is good. 

However it does not follow the coding "style" of g3log.  I do think the functionality change is good, if you want to add them to the g3log please consider the following. 

1. Changing the default timestamp "provider" should be done before the g3log is initialized. 
2. The default timestamp provider can be a lambda or ```std::function<...>``` that you can override
3. The override of the timestamp provider would simply add whatever ```std::function<...>``` or ```lambda``` that can give more exact or different timestamp. 

---> This is pretty much exactly how signal can crashhandling is overriden and changed in g3log today. 
ref: ```g3log.cpp :: g_fatal_to_g3logworker_function_ptr```

src/g3log/logmessage.hpp
gimesketvirtadieni: _timestamp can be either a pointer (raw / unique / shared) or a template because different TimestampProvider's return _timestamp's of different types.

As you pointed out smart pointer brings a bit of overhead during runtime. Template will have impact for compile time, which I think is less critical. I am not sure whether raw pointers will cause cache misses.

So which approach do you think is more suitable in this case?

gimesketvirtadieni:
I will digest your suggestion and get back ;)

src/g3log/logmessage.hpp
KjellKod: Not sure I would pick either, ref: design discussion

gimesketvirtadieni: I checked your solution with g_fatal_to_g3logworker_function_ptr global pointer. By default it points to pushFatalMessageToLogger. This is just an action that does not return any objects.

TimestampProvider does return an object so it is a different beast compared to your g_fatal_to_g3logworker_function_ptr. Moreover, different TimestampProvider's generate timestamps of different types (default provider returns std::time_t, in my case it is boost::posix_time::ptime), which is the the whole point of TimestampProvider API. Because timestamp objects can be of different types there are just two ways of saving them in the log messages:
-> Using pointers (this is what I did and it may have runtime performance impact)
-> Using templates (this is what can be done and it will have compile time impact, which I think is less painful)

Resolving how timestamps are saved in the relevant log message will lead to resolving other items you described below.

KjellKod: What I meant was to have a function-ptr that returned a std::string formatted as desired.

The function-ptr should be called at the creation of the object. And the "time string" would be saved in the message object. I.e it would work with std::time as well as boost::...ptime 

However, I realize it would break the API. The logging sinks should be able to request the time string formatt d in any way possible
Example:
YYYY
YY:SS
YYYY:MM:DD HH:MM:SS

These three are all valid string formats of the time that a sink could use.

---

KjellKod: Further. See my "broken API" comment below.

I will take another look and ponder this. I am not sure how the templated version would work. The sinks should be ignorant over the time type.

At worst this will not be merged and you create your own fork of g3log with a high precision timer. I can then happily point to your g3log version when other users need such high precision.
---

KjellKod: Don't know why it became all bold. I will edit my entry later to make it non-bold

gimesketvirtadieni: Actually you gave a great idea, but it is not exactly what you explained :)

But first I did a bit of analysis using templates for timestamps. Even though it is possible however it would be a HUGE g3log change. Using templates will affect pretty much all the layers: macros, LogCapture, LogMessage, moving most of the implementation code to headers... In other words it is not feasible.

This conclusion left me with using pointers. I started thinking about "placement new()" feature which I have never used, so it would be cool to try it out. It is possible to allocate single memory chunk for log message that would contain log message and timestamp of whatever type using "placement new()" feature. So memory wise log message would be a single block which would prevent from cache misses. Application wise it would be a raw pointers used in log message objects to the timestamps with a bit of woodoo code while creating objects. I almost started playing with this approach but then you proposed using strings...

Frankly I think that using strings in log messages is not a good solution. It is not elegant, it requires reserving some predefined buffer in log message which limits formatting functionality and ... it is not elegant :) However timestamp provider can return a struct with fields year, month, day, hour, minute, sec, microsecond. This way timestamp instance (which by the way would be always of the fixed size which is good for HW) would be part of the log message without any templates, pointers, heap allocation...Formatting would take care of converting this to a string depending of given format on logger's thread.

Would you accept my changes based on this approach?

KjellKod: It does sound more promising. Please consider also the case of systems using the default std:: time and chrono system.

I.e the finest granularity on most systems is second. The ms and us can then just be zeroed.

KjellKod: I would then advice on the function pointer method I suggested earlier. With the difference that it would return your time struct which points to the time of the call into the function. 

I.e I envision 

1) the timestamp provider class to not be needed.
2) the time struct can be stored inside the LogMessage struct
3) the LogMessage struct would provide with similar API as today a way of formatting the time into a string
4) possibly using the posix time struct (why reinvent the wheel?)

Does the above make sense?

gimesketvirtadieni: Yep, it makes sense :)
I will commit changes based on this approach then.

gimesketvirtadieni: Hi

I committed the first draft version based on the approach we discussed with some deviations (bare with me...)

- Instead of std::tm-like struct I kept existing std::time_t so default timestamp initialization remained the same; the reason is to minimize timestamp capture effort, because in case of tm-like struct there is a need to extra conversion std::time_t -> tm struct; besides, keeping std::time_t makes timestamp kind of compatible with chrono
- Then I introduced microsec field in timestamp struct which I plan to replace with boost approach - using num_fractional_digits and fractional_seconds fields instead
- There are two constructors in timestamps: a default one using g3::systemtime_now and a new one which allows providing microsec field (to be changed to num_fractional_digits and fractional_seconds)

It works and I can provide a custom function/functor/lambda to generate timestamps with fractional part (C++14 syntax):
	g3::setTimestampGenerator([format = std::string("%Y/%m/%d %H:%M:%S")]() mutable {
		auto ptime = boost::posix_time::microsec_clock::local_time();

		return g3::Timestamp(
			boost::posix_time::to_time_t(ptime),
			ptime.time_of_day().fractional_seconds(),
			format);
	});

Now I need to figure out formatting part. The straight forward way would be changing localtime_formatted's signature to accept timestamp object instead of std::time_t

Do you agree so far?

src/g3log/logmessage.hpp
```diff
@@ -45,8 +46,8 @@ namespace g3 {
       }
 
       /// use a different format string to get a different look on the time.
-      //  default look is Y/M/D H:M:S
-      std::string timestamp(const std::string& time_format = {internal::date_formatted + " " + internal::time_formatted}) const;
+      //  default look (Y/M/D H:M:S) is defined by default timestamp generator
+      std::string timestamp(const std::string& time_format = "") const;
```
KjellKod: I can buy that we skip the default format argument and instead make it explicit. In that case I would remove the default argument altogether.

src/g3log/logmessage.hpp
```diff
@@ -115,6 +116,9 @@ namespace g3 {
    };
 
 
+   // prototypes for timestamp generator's functionality
+   std::function<Timestamp()>& getTimestampGenerator();
+   void                        setTimestampGenerator(std::function<Timestamp()>);
```
KjellKod: please fix the funky formatting

src/timestamp.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+	Timestamp::~Timestamp() {
+	}
+
+
+	std::string Timestamp::getLocalTimestamp(const std::string& format) const {
+		std::ostringstream stringStream;
+		stringStream << _microsec;
+
+		return g3::localtime_formatted(_time, (format.empty() ? _defaultFormat : format)) + "|" + stringStream.str() + "|";
```
KjellKod: by removing the empty defalt and making it explicit it makes this logic easier. Please see my earlier review

src/timestamp.cpp
```diff
@@ -0,0 +1,53 @@
+#include "g3log/time.hpp"
+#include "g3log/timestamp.hpp"
+#include <sstream>
+
+
+namespace g3 {
+
+
+	Timestamp::Timestamp(std::string& defaultFormat) :
```
KjellKod: This timestamp struct is overcomplicating it and basically re-invents something that already exists.

What reasons are there to not use the already existing `timeval struct`?  As far as I understand it already exists on Windows and Linux. 

Ref: 
1) http://souptonuts.sourceforge.net/code/gettimeofday.c.html
2) https://msdn.microsoft.com/en-us/library/windows/desktop/ms740560(v=vs.85).aspx

With this functionality you can get the `std::time_t ` and microsecond precision which is what you want, right?

If the `timeval struct` works great then I  see no reason for the `TimeStamp` class and the generator.   In that case I think you can simplify this whole pull request to be much less changes. 

A) use the timeval struct and remove the std::time_t part from the `g3::logmessage`
B) use the formatting exactly as before (revert your changes). Just when the std::string formatted string is needed then the `timeval struct` can be converted to a `std::time_t` and then formatted as before. 
C) use the` timeval struct`'s microsecond precision instead of the g3 "microsecond" counter that you didn't like. 
D) remove the microsecond counter.

KjellKod: I.e. it's less "dynamic" than before but I think the code change is way less intrusive and overall cleaner

gimesketvirtadieni: First things first – let’s nail down the data structure.

It is great that you pointed at timeval struct, I was not aware about it. It is exactly what I need: just two fields – time_t and microsecond part. I did a bit of a reading about sys/time.h (the header defining this structure) and it is not part of C++ standard library, but rather part of the POSIX standards:
http://stackoverflow.com/questions/25615571/how-to-use-sys-time-h-in-window
https://www.codefull.org/2015/12/systime-h-replacement-for-windows/

Yes, it means Windows support is limited. I checked your link and as I read it this struct is defined in weird Winsock2.h header which makes me think that functions that operate with this struct are totally different compared to POSIX.

So I think if you decide to go with timeval and sys/time.h then you are sacrificing Windows support (but I am not a Widows expert so I may be wrong).

So should it be custom Timestamp structure with time_t and factional part or should I use POSIX sys/time.h?

KjellKod: Use the posix struct. I'll do some digging. If support on Windows is bad then we at worst have to #ifdef it at: 
It's creation, the formatting function and the struct member itself (Windows : time_t, all other struct timeval)

KjellKod: Use the posix struct.  On the windows side the winsock or windows.h will provide what is needed.

When you get this closer to finish I can provide the code needed to make it work on Windows

gimesketvirtadieni: Sure, will do :)

Quick update: I started implementing changes based on using timeval struct. This struct should be set by gettimeofday. I got curious about the second parameter of this function and clarified that it is depricated:
_"POSIX.1-2008 marks gettimeofday() as obsolete, recommending the use of clock_gettime(2) instead"_

So I looked at clock_gettime which uses timespec struct:
- It is very similar to timeval except that it uses fractional part at nanosec precision
- It part of time.h or ctime header which makes me think it is supported on Windows

I made changes (still very raw) to use timespec and so far it works as expected. Please let me know what you think ;)

KjellKod: I like it.

gimesketvirtadieni: Update: I made quick and dirty comparison of timestamp fractional part generated by clock_gettime(CLOCK_REALTIME, &_timestamp) with timestamps generated by chrono library:
std::chrono::high_resolution_clock::now().time_since_epoch().count();
And I got consistant results :)

So I will just refine timestamp generation code and enhance formating functionality to support fractional part.

src/g3log/logmessage.hpp
```diff
@@ -87,7 +88,7 @@ namespace g3 {
       // Complete access to the raw data in case the helper functions above
       // are not enough.
       //
-      std::time_t _timestamp;
+      Timestamp _timestamp;
```
KjellKod: This is going away right? In favor of the timeval struct

src/logmessage.cpp
```diff
@@ -136,14 +140,15 @@ namespace g3 {
...
-      , _call_thread_id(std::this_thread::get_id())
+      : _call_thread_id(std::this_thread::get_id())
       , _microseconds(microsecondsCounter())
       , _file(splitFileName(file))
       , _line(line)
       , _function(function)
       , _level(level)
-   {}
+   {
+      clock_gettime(CLOCK_REALTIME, &_timestamp);
```
KjellKod: Since c++11 I think you can use timespec_get(...) on both Linux and Windows:

http://en.cppreference.com/w/c/chrono/timespec_get

src/logmessage.cpp
```diff
@@ -122,7 +122,11 @@ namespace g3 {
 
 
    std::string LogMessage::timestamp(const std::string& time_look) const {
-      return  localtime_formatted(_timestamp, time_look);
+      std::ostringstream stringStream;
+
+      stringStream << _timestamp.tv_nsec;
+
+      return g3::localtime_formatted(_timestamp.tv_sec, (time_look.empty() ? g3::internal::date_formatted + " " + g3::internal::time_formatted : time_look)) + "|" + stringStream.str() + "|";
```
KjellKod: Please clean this up

1) The checking of empty is basically a work around to avoid having to have a default formatting. Let's put back the default formatting and instead the empty check can go away

2) I don't understand the "I" part. Please explain.

src/g3log/logmessage.hpp
KjellKod: On second thought. Let's not. Please see below. It becomes quite confusing with the empty check on the formatting.

KjellKod: Please update the comment. There is no timestamp generator

src/logmessage.cpp
```diff
@@ -136,14 +140,15 @@ namespace g3 {
 
    LogMessage::LogMessage(const std::string& file, const int line,
                           const std::string& function, const LEVELS& level)
-      : _timestamp(g3::systemtime_now())
-      , _call_thread_id(std::this_thread::get_id())
+      : _call_thread_id(std::this_thread::get_id())
       , _microseconds(microsecondsCounter())
```
KjellKod: This can be removed now. With this pull request the micro seconds counter becomes obsolete, right?

please remove the g_start_time in this file

src/g3log/logmessage.hpp
gimesketvirtadieni: That's right. Custom Timestamp is gone in the my last commit:
`timespec _timestamp;`

src/logmessage.cpp
gimesketvirtadieni: Thanks for the tip, I will try it out ;)

src/logmessage.cpp
gimesketvirtadieni: I tried to avoid touching formatting part including localtime_formatted function before we sort out data structure and generation step. localtime_formatted accepts only seconds part so I appended fractional part between pipes just to print it out to sinks.
Now data structure and generation are sorted so I will clean up formatting part ;)

src/logmessage.cpp
gimesketvirtadieni: I agree, microsec counter is useless after this request is complete. I will remove it.

src/logmessage.cpp
gimesketvirtadieni: I just tried timespec_get and result is equivalent to clock_gettime's result. So changed to timespec_get

src/logmessage.cpp
KjellKod: Thanks. Please comment to let me know when I should take another look.

It's starting to look finished :)

gimesketvirtadieni: Indeed it is getting better.
I just committed few changes to wire timespec timestamp with formatting part; now put_time receives additional parameter const long nsec which can be used for generating formatted string with fractional sec part (it is still missing though).
Few questions:
- I created one more localtime_formatted function with different signature that accepts timespec argument. Do you want to keep the old one with time_t parameter?
- I think systemtime_now should be changed to return timespec instead of time_t. Do you agree?

If you agree to remove old version of localtime_formatted and change systemtime_now then these will change g3 API as sinks need to be changed accordingly. So it is up to you ;)
- Even though I can already get access to fractional part from sinks like following:
`logMessage.timestamp({"%Y/%m/%d %H:%M:%S"}) + "." + logMessage._timestamp.tv_nsec`
I plan to enhance put_time to support following formatting options: %f - sec fractional part (default nanosec precision); %f[1...6] - sec fraction part (precision is as defined by the digit after %f)

Do you agree?

gimesketvirtadieni: I can see that travis CI is failing to recognize TIME_UTC macro, so I reverted back to clock_gettime

src/time.cpp
```diff
@@ -21,7 +21,7 @@ namespace g3 {
       // This is needed since latest version (at time of writing) of gcc4.7 does not implement this library function yet.
       // return value is SIMPLIFIED to only return a std::string
 
-      std::string put_time(const struct tm *tmb, const char *c_time_format) {
+      std::string put_time(const struct tm *tmb, const long nsec, const char *c_time_format) {
```
KjellKod: The nsec parameter is not used. The reason for overloading std::put_time was that it was not implemented across platforms at the time of writing. It might be the function can be removed in another pull request but either way the API must not be changed

Ref: http://en.cppreference.com/w/cpp/io/manip/put_time

src/logmessage.cpp
KjellKod: I think it might be tricky to provide the formatting together with the nanoseconds. Let's keep the first go at this as simple as possible. 

For now,   I would keep it as it was (time_t argument) and provide a few simple functions instead that can be used by the sinks and of course by the default sink.

1. In logmessage .h/cpp  just a wrapper to return std::time_t from the internal time representation (timespec). I.e it can be converted easily on the Linux side and I can later provide the code for converting it on Windows side.

2. In logmessage .h/cpp a simple, convenience, function to return the nanosecond part as a) value 
size_t nanosecond() {...}


Then it is trivial for the default sink as well as other sinks to continue using exactly the same API as before but with the option of adding the nanoseconds information. 

If a sink wants to format his in a strange funky way (hour / nanosecond / second / year) or whatnot then the writer of that creative sink can deal with that mess.

KjellKod: I did not find any documentation regarding any TIME_UTC macro in cpp reference

gimesketvirtadieni:
Hi

I've been quite busy lately but now have some spare time to complete this pull request.

I have reverted g3 API changes so now put_time and localtime_formatted are as before. I did enhance the overloaded instance of localtime_formatted (the one that accepts timespec) to support %f in the format string. It is not that tricky as I expected ;)

Now sinks can use nansec like following: `logMessage.timestamp({"%Y/%m/%d %H:%M:%S.%f"})`

Let me know what you think ;)

KjellKod:
A quick question. On your platform does the nanosecond granularity give you any actual improved information compared to microsecond?

Or are the nanosecond rounded to nearest microsecond?

src/time.cpp
```diff
@@ -69,6 +71,27 @@ namespace g3 {
    /// * format string must conform to std::put_time
    /// This is similar to std::put_time(std::localtime(std::time_t*), time_format.c_str());
 
+   std::string localtime_formatted(const timespec &time_snapshot, const std::string &time_format) {
+      ushort fcount = 0;
+      for(auto found_ptr = time_format.c_str(); (found_ptr = strstr(found_ptr, "%f")) != nullptr; found_ptr++, fcount++);
+      const size_t format_size = time_format.length() + fcount * 7 + 1;  // len(format) + num(nanosec size - len("%f")) + zero
```
KjellKod: I will pull down your changes and try them.

What does the "f" stand for?  Ref: http://en.cppreference.com/w/cpp/io/manip/put_time

since you  are expanding the API for the format string you can make it obvious (more so than the std)
Why not spelling it out? 
%NS 
Or 
 %NanoSec that way it is ensured to avoid API clashes

src/time.cpp
```diff
@@ -69,6 +71,27 @@ namespace g3 {
...
+      char format_buffer[format_size];
+      strcpy(format_buffer, time_format.c_str());
+
+      for(auto found_ptr = format_buffer; (found_ptr = strstr(found_ptr, "%f")) != nullptr; found_ptr++) {
+         strcpy(found_ptr + 9, found_ptr + 2);
+         memcpy(found_ptr, "000000000", 9);
+         auto nsec_str = std::to_string(time_snapshot.tv_nsec);
+         auto nsec_len = strlen(nsec_str.c_str());
+         memcpy(found_ptr, nsec_str.c_str() + nsec_len - 9, nsec_len);
+      }
```
KjellKod: This all looks correct but it can be simplified a great deal. Something like this should work:

```
std::string localtime_formatted(const timespec &time_snapshot, const std::string &time_format) {
   auto format_buffer = time_format;
   auto pos = format_buffer.find(kNanoSecondIdentifier);
   while (pos != std::string::npos) {
   	  format_buffer.replace(pos, kNanoSecondIdentifier.size(), ns);
   	  pos = format_buffer.find(kNanoSecondIdentifier, pos + ns.size());
   }
      std::tm t = localtime(time_snapshot.tv_sec);
      return g3::internal::put_time(&t, format_buffer);
}
```

ref: simple testing of this at: <script src="http://ideone.com/e.js/lSbs1Z" type="text/javascript" ></script>

KjellKod: FYI: the std way is also faster. ref: http://ideone.com/XNJRkI

gimesketvirtadieni:
Hi

I thought it is hard to bit memcpy :) I changed format string generation part based on your approach (could not stand from adding few things on my own). Please take a look at localtime_formatted and let me know what you think.

src/time.cpp
gimesketvirtadieni: "f" stands for fractional part of a sec. It is used in boost ;) I am fine with any key for fractional part as long as it is possible to define Milli, Micro and Nano sec. If I had to choose I would use: %ML, %MC, %NN.

gimesketvirtadieni: My target platform is cheap ARM boards (Pi and alike) with Linux RT kernel. As far as I know it uses free running counter at 1 MHh. Which means precision granularity is microsec and there is no point in nanosec for me, but it might be different for other g3log users.

src/time.cpp
```diff
@@ -69,6 +71,28 @@ namespace g3 {
...
+      std::string kNanoSecondIdentifier = "%f";
+   }
+
+   std::string localtime_formatted(const timespec &time_snapshot, const std::string &time_format) {
+      auto format_buffer = time_format;
+      auto value_str = std::to_string(time_snapshot.tv_nsec);
+
+      // creating nsec string with leading zeros
+      auto nsec_str = std::string("000000000");
+      nsec_str.replace(nsec_str.length() - value_str.length(), value_str.length(), value_str);
```
KjellKod: I have no particular opinion whether or not leading zeroes is a good thing or not. 

You can simplify this, potentially speed it up?  by creating one string less
```
std::string ns = std::string(number_zeroes - ns_value.length(), '0') + ns_value;
```
I.e. pretty much what you already wrote, only simplified.

KjellKod: ref: http://ideone.com/VUgb9z

src/g3log/logmessage.hpp
```diff
@@ -87,7 +87,7 @@ namespace g3 {
       // Complete access to the raw data in case the helper functions above
       // are not enough.
       //
-      std::time_t _timestamp;
+      timespec _timestamp;
       std::thread::id _call_thread_id;
       int64_t _microseconds;
```
KjellKod: ```int64_t _microseconds;```
this has to be removed

KjellKod: which means changing the logmessage.cpp too (as we discussed previously)

src/g3log/time.hpp
```diff
@@ -45,6 +45,7 @@ namespace g3
    * WARNING: At time of writing there is only so-so compiler support for
    * std::put_time. A possible fix if your c++11 library is not updated is to
    * modify this to use std::strftime instead */
+   std::string localtime_formatted(const timespec &time_snapshot, const std::string &time_format) ;
```
KjellKod: I think this is a good idea. Internally the other localtime_formatted will be used but a client could use it if they wanted to. 

When we set the spec for the formatting flag we should document that in this header too since it is an override of the normal formatting for ```std::put_time```

src/time.cpp
```diff
@@ -34,6 +35,7 @@ namespace g3 {
          //                    ... also ... This is way more buffer space then we need
 
          auto success = std::strftime(buffer, size, c_time_format, tmb);
+         // TODO: incorrect code
```
KjellKod: ???

src/time.cpp
```diff
@@ -69,6 +71,28 @@ namespace g3 {
    /// * format string must conform to std::put_time
    /// This is similar to std::put_time(std::localtime(std::time_t*), time_format.c_str());
 
+   namespace {
+      std::string kNanoSecondIdentifier = "%f";
```
KjellKod: OK. I buy the ```%f``` since it will relate to many people already familiar with boost
http://www.boost.org/doc/libs/1_35_0/doc/html/date_time/date_time_io.html

src/time.cpp
gimesketvirtadieni: // TODO: incorrect code
         if (0 == success)
         {
            // assert is always false because of if condition
            assert((0 != success) && "strftime fails with illegal formatting");

src/g3log/logmessage.hpp
gimesketvirtadieni: Have just committed relevant changes

src/time.cpp
gimesketvirtadieni: Made this simplification in my last commit

src/time.cpp
gimesketvirtadieni: What about Milli / Micro / Nano formatting? Should I use following formatting?:
- %f - default which is nano
- %f3 - millisec
- %f6 - microsec
- %f9 = %f

KjellKod: Sounds good.

src/time.cpp
KjellKod: It's not incorrect but perhaps confusing.

So if the library is built in Debug then for 0 == success then
The assert will always fire. When it fires the assert condition is given (std out/err).
I.e the check in the assert statement is there for the coder more than an actual 
"Real check".

If the library is built in Production then an assert statement is *ignored*. But 
C_time_format is returned instead of the buffer.

I.e the logic is correct

src/time.cpp
KjellKod: I don't see it

gimesketvirtadieni:
Hi

Made the first cut of formatting routine that supports %f[3|6|9] key (it requires unit tests though). Please take a look.

src/time.cpp
```diff
@@ -69,6 +72,51 @@ namespace g3 {
...
+         char ch = (format_buffer.size() > pos + kIdentifier.size() ? format_buffer.at(pos + kIdentifier.size()) : '\0');
+         switch (ch) {
+            case '3': index = 0;    break;
+            case '6': index = 1;    break;
+            case '9': index = 2;    break;
+	    default : ch    = '\0'; break;
+         }
+
+         // replacing %f with sec fractional part value
+         format_buffer.replace(pos, kIdentifier.size() + (ch == '\0' ? 0 : 1), value_str[index]);
```
KjellKod: I think it overall looks good. 

I would advice on having the cost for the string creation only be done if it is needed. I.e now each log entry pays the penalty for creating the three fractional strings even if it's not needed. 

The more remote chance that someone wants all fractions of a second should be considered the exceptional case and in that case that use case can pay for the extra overhead.

I.e.  som pseudo code

```
enum class TimeFractions {Milli, Micro, Nano};
```
something like ```
         switch (ch) {
             case '3': fraction = TimeFractions::Milli;    break;
             case '6': fraction = TimeFractions::Micro;    break;
             case '9': fraction = TimeFractions::Nano;    break;
 	    default : ch    = '\0'; break;
          }
```

Then
```
if (fraction != TimeFractions::Ignored) {
   format_buffer.replace(pos,  fraction_identifier_size(fraction), fraction_tostring(fraction,  time_snapshot));
}
```

And 
std::string  fraction_tostring(TimeFractions fraction,  const timespec& time_snapshot) {
   switch (fraction) {
     case TimeFractions::Milli: return /*....*/
 .... 
}
```

KjellKod: I.e. 

* Break apart the function so it is less complex and can be unit tested sensibly. (TDD?)
* Change the function so that the cost mostly for when actual fractional usage is needed

src/time.cpp
```diff
@@ -69,6 +72,51 @@ namespace g3 {
...
+      for (ushort i = 0; i < sizeof(kZeroes)/sizeof(*kZeroes); i++) {
+         value_str[i] = std::string(kZeroes[i] - value_str[i].length(), '0') + value_str[i];
+      }
+
+      // replacing %f[3|6|9] with actual sec fractional value
+      for(size_t pos = 0; (pos = format_buffer.find(kIdentifier, pos)) != std::string::npos; pos += kIdentifier.size()) {
+
+         // figuring out whether this is nano, micro or milli identifier
+         ushort index = 2;
+         char ch = (format_buffer.size() > pos + kIdentifier.size() ? format_buffer.at(pos + kIdentifier.size()) : '\0');
```
KjellKod: I would split this out to its own function so it can also be easily tested.
```
TimeFractions fraction = fraction_type(format_buffer, pos);
```

_See below and it makes sense_

src/time.cpp
```diff
@@ -69,6 +72,51 @@ namespace g3 {
...
+         switch (ch) {
```
KjellKod: I would split this out to its own function so it can also be easily tested.
```
TimeFractions fraction = fraction_type(format_buffer, pos);
```

_See the comment below and it makes sense_

src/time.cpp
gimesketvirtadieni: I had the same thinking: normally people will be interested in either of fraction type. I moved fraction-2-string conversion inside the search loop.
If you are fine with the algorithm now then I can look into splitting it into smaller functions and providing unit tests.

KjellKod: Sounds good. Go ahead.

I tested it on OSX and I will try to come with an update to you (pull request to your pull request)

gimesketvirtadieni:
I refined localtime_formatted; introduced two help functionas to make localtime_formatted cleaner. Now I am learning gtest (I am JAVA guys so I am used to JUnit ;) to write unit tests covering the new localtime_formatted.

src/logmessage.cpp
```diff
@@ -136,14 +125,16 @@ namespace g3 {
...
       , _file(splitFileName(file))
       , _line(line)
       , _function(function)
       , _level(level)
-   {}
+   {
+      // Falling back to clock_gettime as TIME_UTC is not recognized by travis CI
+      // timespec_get(&_timestamp, TIME_UTC);
+      clock_gettime(CLOCK_REALTIME, &_timestamp);
+   }
```
KjellKod: // TODO --- KjellKod to supply ```clock_gettime``` overloads for OSX and Windows

KjellKod:
@gimesketvirtadieni thanks I think this pretty much nailed it. 

You don't have to take care of the unit tests, I can do that since I have to add support for OSX and Windows at the same time. 

Unless you have left something out I suggest we close this pull request and I can continue on another pull request that inherits this one.   _All your commits will of course show up in the git history for the finished code._ 

ref: https://github.com/KjellKod/g3log/pull/104

PR #96, opened 2016-05-28
added automatic linking pragma for `StackWalk64`

d-led:
this simplifies build configs for windows

### DIFF
--- a/src/stacktrace_windows.cpp
+++ b/src/stacktrace_windows.cpp
@@ -23,6 +23,8 @@
 #include <mutex>
 #include <g3log/g3log.hpp>
 
+#pragma comment(lib, "dbghelp.lib")
+
 
 #if !(defined(WIN32) || defined(_WIN32) || defined(__WIN32__))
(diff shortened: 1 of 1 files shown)


src/stacktrace_windows.cpp
```diff
@@ -23,6 +23,8 @@
 #include <mutex>
 #include <g3log/g3log.hpp>
 
+#pragma comment(lib, "dbghelp.lib")
```
KjellKod: Can we just Link with dbghelp.lib?

Ref: https://github.com/KjellKod/g3sinks/blob/master/logrotate/test/Test.cmake

I don't have access to Windows so I can't test either. But I figured if it is enough to just link it then for Windows users it would a done deal?

d-led: Yes, but this is more or less the windows idiomatic way. Boost relies on the same mechanism. These standard windows libraries are sometimes hard to chase down, which one to link. But, of course, if the documentation and the build config say what to link, it's not strictly necessary.

KjellKod:
:+1:

PR #98, opened 2016-06-12
fix to implicit type casting

jhlee8804:
implicit type casting to static casting


jhlee8804:
some commits (e321639, 2403669) are my mistake.

KjellKod:
:+1:

PR #101, opened 2016-06-22
Fixed a bug causing cross-compilers to fail

Nitaym:
(The edited line eliminated the original CXXFLAGS variable, removing the sysroots parameter, which is needed for cross compiling)


KjellKod:
:+1:

PR #105, opened 2016-07-06
Added "removeSink" method and related functionality to LogWorker API

gimesketvirtadieni:
This is a pull request for dynamic sink removal functionality

### DIFF
--- a/src/g3log/logworker.hpp
+++ b/src/g3log/logworker.hpp
@@ -109,14 +109,6 @@ namespace g3 {
          removeWrappedSink(sink_handle->sink());
       }
 
-      void flush(bool propogate_to_sinks = true) {
-         g3::spawn_task([propogate_to_sinks] {
-            if (propogate_to_sinks) {
-               // TODO: ...
-            }
-         }, _impl._bg.get()).wait();
-      }
-
       /// internal:
       /// pushes in background thread (asynchronously) input messages to log file
       void save(LogMessagePtr entry);
--- a/src/g3log/logworker.hpp
+++ b/src/g3log/logworker.hpp
@@ -53,6 +53,7 @@ namespace g3 {
    class LogWorker final {
       LogWorker() = default;
       void addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> wrapper);
+      void removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> wrapper);
 
       LogWorkerImpl _impl;
       LogWorker(const LogWorker&) = delete;
@@ -101,6 +102,12 @@ namespace g3 {
          return std2::make_unique<SinkHandle<T>> (sink);
       }
 
+      /// Removes a sink
+      /// @param
+      template<typename T>
+      void removeSink(std::unique_ptr<SinkHandle<T>> sink_handle) {
+         removeWrappedSink(sink_handle.get()->sink());
+      }
 
 
       /// internal:
--- a/src/g3log/logworker.hpp
+++ b/src/g3log/logworker.hpp
@@ -109,6 +109,13 @@ namespace g3 {
          removeWrappedSink(sink_handle->sink());
       }
 
+      void flush(bool propogate_to_sinks = true) {
+         g3::spawn_task([propogate_to_sinks] {
+            if (propogate_to_sinks) {
+               // TODO: ...
+            }
+         }, _impl._bg.get()).wait();
+      }
 
       /// internal:
       /// pushes in background thread (asynchronously) input messages to log file
--- a/src/g3log/logworker.hpp
+++ b/src/g3log/logworker.hpp
@@ -106,7 +106,7 @@ namespace g3 {
       /// @param
       template<typename T>
       void removeSink(std::unique_ptr<SinkHandle<T>> sink_handle) {
-         removeWrappedSink(sink_handle.get()->sink());
+         removeWrappedSink(sink_handle->sink());
       }
 
 
--- a/src/g3log/sinkhandle.hpp
+++ b/src/g3log/sinkhandle.hpp
@@ -24,6 +24,7 @@ namespace g3 {
    // in the resulting future. Ref: SinkHandle::call
    template<class T>
    class SinkHandle {
+      friend class LogWorker;
       std::weak_ptr<internal::Sink<T>> _sink;
 
    public:
@@ -47,6 +48,11 @@ namespace g3 {
             promise.set_exception(std::make_exception_ptr(e));
             return std::move(promise.get_future());
          }
+      }
+
+   protected:
+      std::shared_ptr<internal::Sink<T>> sink() {
+         return _sink.lock();
       }
    };
 }
--- a/src/g3log/sinkhandle.hpp
+++ b/src/g3log/sinkhandle.hpp
@@ -24,7 +24,6 @@ namespace g3 {
    // in the resulting future. Ref: SinkHandle::call
    template<class T>
    class SinkHandle {
-      friend class LogWorker;
       std::weak_ptr<internal::Sink<T>> _sink;
 
    public:
@@ -50,7 +49,6 @@ namespace g3 {
          }
       }
 
-   protected:
       std::shared_ptr<internal::Sink<T>> sink() {
          return _sink.lock();
       }
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -113,9 +113,8 @@ namespace g3 {
    }
 
    void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_addsink_call = [this, &sink] {_impl._sinks.push_back(sink);};
-      auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
-      token_done.wait();
+      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
+      g3::spawn_task(bg_addsink_call, _impl._bg.get());
    }
 
    std::unique_ptr<LogWorker> LogWorker::createLogWorker() {
@@ -123,11 +122,13 @@ namespace g3 {
    }
 
    void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_removesink_call = [this, &sink] {
+      auto bg_removesink_call = [this, sink] {
          _impl._sinks.erase(std::remove(_impl._sinks.begin(), _impl._sinks.end(), sink), _impl._sinks.end());
+
+         // TODO: to be added to the sink API
+         //sink->flush();
       };
-      auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
-      token_done.wait();
+      g3::spawn_task(bg_removesink_call, _impl._bg.get());
    }
 
    std::unique_ptr<FileSinkHandle>LogWorker::addDefaultLogger(const std::string& log_prefix, const std::string& log_directory, const std::string& default_id) {
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -113,8 +113,16 @@ namespace g3 {
    }
 
    void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
-      g3::spawn_task(bg_addsink_call, _impl._bg.get());
+
+      // capturing sink by reference is safe because this method waits for the completion of this task
+      // passing sink by value may lead to keeping it in other thread's context even after this method returns
+      auto bg_addsink_call = [this, &sink] {
+         _impl._sinks.push_back(sink);
+      };
+      auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
+
+      // even though add sink can use async semantic, making it sync improves overall performance
+      token_done.wait();
    }
 
    std::unique_ptr<LogWorker> LogWorker::createLogWorker() {
@@ -122,13 +130,19 @@ namespace g3 {
    }
 
    void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_removesink_call = [this, sink] {
-         _impl._sinks.erase(std::remove(_impl._sinks.begin(), _impl._sinks.end(), sink), _impl._sinks.end());
 
-         // TODO: to be added to the sink API
-         //sink->flush();
+      // capturing sink by reference is safe because this method waits for this task to complete
+      // passing sink by value may lead to keeping it in other thread's context even after this method returns
+      auto bg_removesink_call = [this, &sink] {
+
+         // once sink is removed from worker, it will not receive new log entries so its destructor can flush sink's content safely
+         // sink flush is done implicitly dy destroying as capturing sink by reference and waiting for task to complete guarantees sink is destroyed
+         _impl._sinks.erase(std::remove(_impl._sinks.begin(), _impl._sinks.end(), sink), _impl._sinks.end());
       };
-      g3::spawn_task(bg_removesink_call, _impl._bg.get());
+      auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
+
+      // even though remove sink can use async semantic, making it sync improves overall performance
+      token_done.wait();
    }
 
    std::unique_ptr<FileSinkHandle>LogWorker::addDefaultLogger(const std::string& log_prefix, const std::string& log_directory, const std::string& default_id) {
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -113,7 +113,7 @@ namespace g3 {
    }
 
    void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
+      auto bg_addsink_call = [this, &sink] {_impl._sinks.push_back(sink);};
       auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
       token_done.wait();
    }
@@ -123,15 +123,10 @@ namespace g3 {
    }
 
    void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_removesink_call = [this, sink] {
+      auto bg_removesink_call = [this, &sink] {
          _impl._sinks.erase(std::remove(_impl._sinks.begin(), _impl._sinks.end(), sink), _impl._sinks.end());
       };
       auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
-      token_done.wait();
-
-      // this will ensure captured sink pointer value will be released
-      auto bg_empty_call = [] {};
-      token_done = g3::spawn_task(bg_empty_call, _impl._bg.get());
       token_done.wait();
    }
 
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -29,6 +29,8 @@ namespace g3 {
 
       if (_sinks.empty()) {
          std::string err_msg {"g3logworker has no sinks. Message: ["};
+
+         // TODO: following line will cause seg fault if empty msgPtr is provided
          err_msg.append(uniqueMsg.get()->toString()).append({"]\n"});
          std::cerr << err_msg;
       }
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -123,7 +123,7 @@ namespace g3 {
    void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
       auto bg_removesink_call = [this, sink] {
          for (auto it = _impl._sinks.begin(); it != _impl._sinks.end(); ) {
-            if ((*it).get() == sink.get()) {
+            if ((*it) == sink) {
                it = _impl._sinks.erase(it);
             } else {
                ++it;
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -120,6 +120,20 @@ namespace g3 {
       return std::unique_ptr<LogWorker>(new LogWorker);
    }
 
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
+      auto bg_removesink_call = [this, sink] {
+         for (auto it = _impl._sinks.begin(); it != _impl._sinks.end(); ) {
+            if ((*it).get() == sink.get()) {
+               it = _impl._sinks.erase(it);
+            } else {
+               ++it;
+            }
+         }
+      };
+      auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
+      token_done.wait();
+   }
+
    std::unique_ptr<FileSinkHandle>LogWorker::addDefaultLogger(const std::string& log_prefix, const std::string& log_directory, const std::string& default_id) {
       return addSink(std2::make_unique<g3::FileSink>(log_prefix, log_directory, default_id), &FileSink::fileWrite);
    }
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -110,8 +110,13 @@ namespace g3 {
       _impl._bg->send([this, fatal_message] {_impl.bgFatal(fatal_message); });
    }
 
-   void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
+   void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sinkPtr) {
+      std::weak_ptr<g3::internal::SinkWrapper> sinkWeakPtr = sinkPtr;
+      auto bg_addsink_call = [this, sinkWeakPtr] {
+         if (auto sinkPtr = sinkWeakPtr.lock()) {
+            _impl._sinks.push_back(sinkPtr);
+         }
+      };
       auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
       token_done.wait();
    }
@@ -120,14 +125,18 @@ namespace g3 {
       return std::unique_ptr<LogWorker>(new LogWorker);
    }
 
-   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_removesink_call = [this, sink] {
-         for (auto it = _impl._sinks.begin(); it != _impl._sinks.end(); ) {
-            if ((*it) == sink) {
-               it = _impl._sinks.erase(it);
-            } else {
-               ++it;
-            }
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sinkPtr) {
+      std::weak_ptr<g3::internal::SinkWrapper> sinkWeakPtr = sinkPtr;
+      auto bg_removesink_call = [this, sinkWeakPtr] {
+         if (auto sinkPtr = sinkWeakPtr.lock()) {
+            _impl._sinks.erase(
+               std::remove(
+                  _impl._sinks.begin(),
+                  _impl._sinks.end(),
+                  sinkPtr
+               ),
+               _impl._sinks.end()
+            );
          }
       };
       auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -110,13 +110,8 @@ namespace g3 {
       _impl._bg->send([this, fatal_message] {_impl.bgFatal(fatal_message); });
    }
 
-   void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sinkPtr) {
-      std::weak_ptr<g3::internal::SinkWrapper> sinkWeakPtr = sinkPtr;
-      auto bg_addsink_call = [this, sinkWeakPtr] {
-         if (auto sinkPtr = sinkWeakPtr.lock()) {
-            _impl._sinks.push_back(sinkPtr);
-         }
-      };
+   void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
+      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
       auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
       token_done.wait();
    }
@@ -125,21 +120,23 @@ namespace g3 {
       return std::unique_ptr<LogWorker>(new LogWorker);
    }
 
-   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sinkPtr) {
-      std::weak_ptr<g3::internal::SinkWrapper> sinkWeakPtr = sinkPtr;
-      auto bg_removesink_call = [this, sinkWeakPtr] {
-         if (auto sinkPtr = sinkWeakPtr.lock()) {
-            _impl._sinks.erase(
-               std::remove(
-                  _impl._sinks.begin(),
-                  _impl._sinks.end(),
-                  sinkPtr
-               ),
-               _impl._sinks.end()
-            );
-         }
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
+      auto bg_removesink_call = [this, sink] {
+         _impl._sinks.erase(
+            std::remove(
+               _impl._sinks.begin(),
+               _impl._sinks.end(),
+               sink
+            ),
+            _impl._sinks.end()
+         );
       };
       auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
+      token_done.wait();
+
+      // this will ensure captured sink pointer value will be released
+      auto bg_empty_call = [] {};
+      token_done = g3::spawn_task(bg_empty_call, _impl._bg.get());
       token_done.wait();
    }
 
--- a/src/logworker.cpp
+++ b/src/logworker.cpp
@@ -124,14 +124,7 @@ namespace g3 {
 
    void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
       auto bg_removesink_call = [this, sink] {
-         _impl._sinks.erase(
-            std::remove(
-               _impl._sinks.begin(),
-               _impl._sinks.end(),
-               sink
-            ),
-            _impl._sinks.end()
-         );
+         _impl._sinks.erase(std::remove(_impl._sinks.begin(), _impl._sinks.end(), sink), _impl._sinks.end());
       };
       auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
       token_done.wait();
--- a/test_unit/test_sink.cpp
+++ b/test_unit/test_sink.cpp
@@ -50,17 +50,19 @@ using namespace g3;
       EXPECT_FALSE(flag->load());
       EXPECT_TRUE(0 == count->load());
 
-      LogMessagePtr message{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
-      message.get()->write().append("this message should trigger an atomic increment at the sink");
-      worker->save(message);
+      LogMessagePtr message1{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
+      message1.get()->write().append("this message should trigger an atomic increment at the sink");
+      worker->save(message1);
+
       worker->removeSink(std::move(handle));
       EXPECT_TRUE(flag->load());
       EXPECT_TRUE(1 == count->load());
 
-      // TODO: saving message after sink was removed will cause seg fault in worker destructor
-      //worker->save(message);
+      LogMessagePtr message2{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
+      message2.get()->write().append("this message is issued after all sinks are removed");
+      worker->save(message2);
    }
-   //EXPECT_TRUE(1 == count->load());
+   EXPECT_TRUE(1 == count->load());
 }
 
 namespace {
--- a/test_unit/test_sink.cpp
+++ b/test_unit/test_sink.cpp
@@ -30,7 +30,6 @@ using namespace g3;
    {
       auto worker = g3::LogWorker::createLogWorker();
       auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
-      worker->flush();
       EXPECT_FALSE(flag->load());
       EXPECT_TRUE(0 == count->load());
       LogMessagePtr message{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
@@ -48,7 +47,6 @@ using namespace g3;
    {
       auto worker = g3::LogWorker::createLogWorker();
       auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
-      worker->flush();
       EXPECT_FALSE(flag->load());
       EXPECT_TRUE(0 == count->load());
 
@@ -57,7 +55,6 @@ using namespace g3;
       worker->save(message1);
 
       worker->removeSink(std::move(handle));
-      worker->flush();
       EXPECT_TRUE(flag->load());
       EXPECT_TRUE(1 == count->load());
 
--- a/test_unit/test_sink.cpp
+++ b/test_unit/test_sink.cpp
@@ -30,6 +30,7 @@ using namespace g3;
    {
       auto worker = g3::LogWorker::createLogWorker();
       auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
+      worker->flush();
       EXPECT_FALSE(flag->load());
       EXPECT_TRUE(0 == count->load());
       LogMessagePtr message{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
@@ -47,6 +48,7 @@ using namespace g3;
    {
       auto worker = g3::LogWorker::createLogWorker();
       auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
+      worker->flush();
       EXPECT_FALSE(flag->load());
       EXPECT_TRUE(0 == count->load());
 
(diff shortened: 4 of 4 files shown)


src/logworker.cpp
```diff
@@ -120,6 +120,20 @@ namespace g3 {
...
+      auto bg_removesink_call = [this, sink] {
+         for (auto it = _impl._sinks.begin(); it != _impl._sinks.end(); ) {
+            if ((*it).get() == sink.get()) {
+               it = _impl._sinks.erase(it);
+            } else {
+               ++it;
+            }
+         }
+      };
+      auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
```
KjellKod: Looks overall very good. Please add unit tests

src/g3log/sinkhandle.hpp
```diff
@@ -24,6 +24,7 @@ namespace g3 {
    // in the resulting future. Ref: SinkHandle::call
    template<class T>
    class SinkHandle {
+      friend class LogWorker;
```
KjellKod: I might be missing something, why is `friend` needed?

src/g3log/logworker.hpp
```diff
@@ -101,6 +102,12 @@ namespace g3 {
          return std2::make_unique<SinkHandle<T>> (sink);
       }
 
+      /// Removes a sink
+      /// @param
+      template<typename T>
+      void removeSink(std::unique_ptr<SinkHandle<T>> sink_handle) {
```
KjellKod: So the `SinkHandle<T>` is received as a client as a `unique_ptr`. Since it is "removed" here it makes sense to "give it up" (never to be seen again). Is that why you chose to send it as a `unique_ptr`?

I.e. the call would be 

```logworker->removeSink(std::move(myHandle));```

File: src/logworker.cpp (hunk 1 above)
Comment (KjellKod) [inline src/logworker.cpp]: I'm not sure if we have to worry about nullptrs. Do you see any cases where a sink handle can be removed and then the sink handle is re-used again but now resulting in that the lock to the sink handle returns nullptr?

Comment (KjellKod) [inline src/logworker.cpp]: Suggestion for unit test. 

Add  a simple log sink where you can access the result of it after it is received at the sink. 
Verify that you see the result when the sink is there. 

Remove the sink
Verify that you cannot see the result when the sink is not there. 

... beware of race ...

Comment (KjellKod) [inline src/logworker.cpp]: ... Another thing that can be tested is that the sink can set a atomic, shared flag when it is destroyed and you can watch that in the test.

File: src/g3log/sinkhandle.hpp (hunk 2 above)
Comment (gimesketvirtadieni) [inline src/g3log/sinkhandle.hpp]: This is required because SinkHandle::sink() method is protected and it is accessed from LogWorker::removeSink(...)

Comment (gimesketvirtadieni) [inline src/g3log/sinkhandle.hpp]: z

File: src/logworker.cpp (hunk 1 above)
Comment (gimesketvirtadieni) [inline src/logworker.cpp]: I am sure what bothers you about nullptr. I am locking weak_ptr which may result to null but then it will be safely handled by shared_ptr == operator comparison (at least that was the idea).
I will provide unit tests for remove sink functionality ;)

Comment (gimesketvirtadieni) [inline src/logworker.cpp]: Both adding and removing sink is done on the same background thread. Message distribution to sinks is done also on the same background thread. So all these operations do not need to be synchronized. Could you elaborate on what race conditions I should be aware of?

File: src/g3log/sinkhandle.hpp (hunk 2 above)
Comment (KjellKod) [inline src/g3log/sinkhandle.hpp]: In that case I'd rather have the `sink()` to be a public function.

File: src/logworker.cpp (hunk 1 above)
Comment (KjellKod) [inline src/logworker.cpp]: Just that depending how you do it, strict unit test, or going through the existing g3 API will mean initiating of  adding / removing sinks might be done in a different thread.  

I.e. when you add a sink then the actual "adding" happens in a different thread then in the client thread that called the logworker API for adding a sink. 

Depending on how you implement the tests it may or may not be an issue.

Comment (gimesketvirtadieni) [inline src/logworker.cpp]: I got it, your comments about nullptr and race conditions were related to unit test ;)

File: src/g3log/sinkhandle.hpp (hunk 2 above)
Comment (gimesketvirtadieni) [inline src/g3log/sinkhandle.hpp]: I made it public in my last commit

File: src/g3log/logworker.hpp (hunk 3 above)
Comment (gimesketvirtadieni) [inline src/g3log/logworker.hpp]: That's correct. addSink returns a handle as a unique_ptr, which means caller is nominated with ownership. By removing the sink caller is giving up ownership and sink can be destroyed.

Comment (KjellKod) [inline src/g3log/logworker.hpp]: Nice!

File: src/logworker.cpp [hunk 4]
```diff
@@ -120,6 +120,20 @@ namespace g3 {
       return std::unique_ptr<LogWorker>(new LogWorker);
    }
 
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
+      auto bg_removesink_call = [this, sink] {
+         for (auto it = _impl._sinks.begin(); it != _impl._sinks.end(); ) {
+            if ((*it) == sink) {
+               it = _impl._sinks.erase(it);
```
KjellKod: You can use the std::find  also.

auto& v = _impl._sinks;
auto it = std::find(v.begin(), v.end(), sink);
if(it != v.end())
    v.erase(it);

gimesketvirtadieni: Yes, I made a change which is similar to your suggestion. Please take a look ;) I also changed the way sink shared_ptr is passed to lambdas in add/removeSink methods. I switched to weak_ptr instead to prevent keeping shared_ptr instances as captured lambda values. This allows making simple and clear test cases for this functionality; otherwise sink cannot be destroyed easily ;)

gimesketvirtadieni:
I started working on unit tests and extended one existing test case to cover removeSink functionality. This is to give you an idea if we are moving in the right direction.

src/logworker.cpp
KjellKod: That doesn't sound right.

Are you saying that the lambda is kept alive past the scope of the function call in which it is executed?

KjellKod:
"
> I switched to weak_ptr instead to prevent keeping shared_ptr instances 
as captured lambda values. This allows making simple and clear test cases 
for this functionality; otherwise sink cannot be destroyed easily
"

That doesn't sound right.

Are you saying that the lambda is kept alive past the scope of the function call in which it is executed?
I.e the lambda with its shared_ptr is kept alive after that the function "addSink" is done with it?

test_unit/test_sink.cpp
```diff
@@ -27,17 +27,21 @@ TEST(Sink, OneSink) {
...
-      EXPECT_FALSE(flag->load());
-      EXPECT_TRUE(0 == count->load());
-      LogMessagePtr message{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
-      message.get()->write().append("this message should trigger an atomic increment at the sink");
-      worker->save(message);
-   }
-   EXPECT_TRUE(flag->load());
+   auto worker = g3::LogWorker::createLogWorker();
+   auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
+   EXPECT_FALSE(flag->load());
```
KjellKod: Please leave the test as it was and create new tests instead.
This test verifies that a sink can be added and it can be verified that the sink can receive data. )

From your tests I expect that 
1. A sink is added
(1b. Optionally verify that the sink can receive data. This could be done with a shared data counter)
2. A sink is removed
3. A flag is set to prove that the sink was destroyed
(4. Optionally verify that once the sink is gone sending log messages will not increment the data counter)

It should be fairly basic since a sink can be any construct you choose.

src/logworker.cpp
```diff
@@ -110,8 +110,13 @@ namespace g3 {
       _impl._bg->send([this, fatal_message] {_impl.bgFatal(fatal_message); });
    }
 
-   void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
+   void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sinkPtr) {
+      std::weak_ptr<g3::internal::SinkWrapper> sinkWeakPtr = sinkPtr;
+      auto bg_addsink_call = [this, sinkWeakPtr] {
+         if (auto sinkPtr = sinkWeakPtr.lock()) {
```
KjellKod: I don't see the reason for this. Are you saying that the lambda kept the shared_ptr alive once this function had gone out of scope? I.e the lambda somehow did not fall out of scope?

src/logworker.cpp
```diff
@@ -120,6 +125,24 @@ namespace g3 {
       return std::unique_ptr<LogWorker>(new LogWorker);
    }
 
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sinkPtr) {
+      std::weak_ptr<g3::internal::SinkWrapper> sinkWeakPtr = sinkPtr;
+      auto bg_removesink_call = [this, sinkWeakPtr] {
+         if (auto sinkPtr = sinkWeakPtr.lock()) {
```
KjellKod: I would advise against the habit of reusing the same name as previously used. That can easily lead to situations where you have hard to find bugs

src/logworker.cpp
KjellKod: Please don't reuse a function member's name. That can lead to hard to find bugs.

gimesketvirtadieni:
When I output number of copies of sinkPtr in the weak_ptr version:
`void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sinkPtr) {
   std::cout << "1 sinkPtr.use_count=" << sinkPtr.use_count() << std::endl;
   std::weak_ptr<g3::internal::SinkWrapper> sinkWeakPtr = sinkPtr;
   auto bg_addsink_call = [this, sinkWeakPtr] {
      if (auto sinkPtr = sinkWeakPtr.lock()) {
         _impl._sinks.push_back(sinkPtr);
      }
   };
   std::cout << "2 sinkPtr.use_count=" << sinkPtr.use_count() << std::endl;
   auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
   token_done.wait();
   std::cout << "3 sinkPtr.use_count=" << sinkPtr.use_count() << std::endl;
}`

I get following out:
1 sinkPtr.use_count=2
2 sinkPtr.use_count=2
3 sinkPtr.use_count=3

The last entry indicates that a new shared_ptr copy was saved in the _sinks vector, which is the expected result.

When I use addSink version based on capturing shared_ptr:
`void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sinkPtr) {
   std::cout << "1 sinkPtr.use_count=" << sinkPtr.use_count() << std::endl;
   auto bg_addsink_call = [this, sinkPtr] {
      _impl._sinks.push_back(sinkPtr);
   };
   std::cout << "2 sinkPtr.use_count=" << sinkPtr.use_count() << std::endl;
   auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
   token_done.wait();
   std::cout << "3 sinkPtr.use_count=" << sinkPtr.use_count() << std::endl;
}`

I get following output:
1 sinkPtr.use_count=2
2 sinkPtr.use_count=3
**3 sinkPtr.use_count=5**

As you can see number of copies indicates that lambda (which captures sinkPtr) was left somewhere along the chain (the same is valid both for add and remove methods). This is due to the fact that lambda is passed by copy multiple times. During the next call variables holding lambda will be overwritten and sinkPtr will be released.

Anyway this is a problem I discovered when tried to run removeSink tests. What do you think about this?

P.S. This is the reason I never capture shared_ptr in lambdas ;)

KjellKod:
Try doing the printout after the scope has finished not inside the still active scope

gimesketvirtadieni:
I did, sinkPtr use_count=3 outside of add sink whereas correct value should be 2: one is local variable and one is in the _sinks vector. If you are not convinced yet that somewhere lambda is left along with sink pointer preventing from correct clean-up, I reverted back to capturing shared_ptr; I also introduced spawning an empty lambda task in the remove sink method. This step clears that something holding previous lambda and ensures expected result.

Looking forward for your ideas

KjellKod:
@gimesketvirtadieni  it is exactly as I said. 

So with the only change to master being:
```
      std::shared_ptr<internal::Sink<T>> sink() {
         return _sink.lock();
      }
```

Then running the following code: 
```
   auto worker = g3::LogWorker::createLogWorker();
   auto handle= worker->addDefaultLogger(argv[0], path_to_log_file);
   auto realSink = handle->sink();
   std::cout << __FUNCTION__ << "number of copies of shared_ptr: " << realSink.use_count() <<    std::endl;
```

I.e. the output of the copies is done when the function call is finished an any intermediate copies of the shared_ptr are deleted. This will output **2** as the number of instances which is what to be expected. 

If I on the other hand put the log output inside the "addWrappedSink" 
```
   void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
      {
         auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
         auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
         token_done.wait();
      }
         std::cout << __FUNCTION__ << ", line: " << __LINE__ << " number of copies of shared_ptr: " << sink.use_count() << std::endl;

   }
```
Then the printout inside the lambda gives **3**.   But as soon (as pointed out above) the as this function exits the scope cleanup will reduce the copies again. (see first example). 

I.e. to the best of my knowledge ```addWrappedSink``` does not have to be changed.

KjellKod:
@gimesketvirtadieni  it is exactly as I said. 

So with the only change to master being:
```
      std::shared_ptr<internal::Sink<T>> sink() {
         return _sink.lock();
      }
```

Then running the following code: 
```
   auto worker = g3::LogWorker::createLogWorker();
   auto handle= worker->addDefaultLogger(argv[0], path_to_log_file);
   auto realSink = handle->sink();
   std::cout << __FUNCTION__ << "number of copies of shared_ptr: " 
                  << realSink.use_count() <<    std::endl;
```

I.e. the output of the copies is done when the function call is finished an any intermediate copies of the shared_ptr are deleted. This will output **2** as the number of instances which is what to be expected. 
One instance is kept in the internal storage and once instance was retrieved by the client ("realSink")

If I on the other hand put the log output inside the "addWrappedSink" 
```
   void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
      {
         auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
         auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
         token_done.wait();
      }
         std::cout << __FUNCTION__ << ", line: " << __LINE__ << " number of copies of shared_ptr: " << sink.use_count() << std::endl;

   }
```
Then the printout inside the lambda gives **3**.   But as soon (as pointed out above) this function exits the scope cleanup will reduce the copies again. (see first example). 

I.e. to the best of my knowledge ```addWrappedSink``` does not have to be changed.

KjellKod:
I can add that I tested this on Clang and Gcc with identical behavior for both.  You are doing yourself a disfavor by printing inside of the `addWappedSink(...)` when the scope isn't cleared.

gimesketvirtadieni:
I think I start to understand where this confusion comes from. Indeed when I try printing out ref count in my main program right after addSink I get 2, which is the correct result. However I face this problem when I run unit tests.

Could you try one simple thing in your setup? Please print out ref count from the first sink unit test right after addSink like following:
TEST(Sink, OneSink) {
...
      auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
      std::cout << "sink.use_count=" << handle->sink().use_count() << std::endl;
...
}
Then please build unit tests and run ./build_travis/test_sink.

What value do you get?

KjellKod:
```
TEST(Sink, OneSink) {
   using namespace g3;
   AtomicBoolPtr flag = make_shared < atomic<bool >> (false);
   AtomicIntPtr count = make_shared < atomic<int >> (0);
   {
      auto worker = g3::LogWorker::createLogWorker();
      auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
      std::cout << "sink use count =" << handle->sink().use_count() << std::endl;
```

running unit tests:
```
kjell.hedstrom@USMAC1541KHEDS build (master *) $ ./test_sink
[==========] Running 8 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 1 test from Sink
[ RUN      ] Sink.OneSink
sink use count =2
[       OK ] Sink.OneSink (103 ms)
[----------] 1 test from Sink (103 ms total)
```

I.e. same result.

gimesketvirtadieni:
Very weird. I do the same thing: take master make two changes (add sink() method and and print out line) and get 3 :(
```
[==========] Running 8 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 1 test from Sink
[ RUN      ] Sink.OneSink
sink use count =3
```
I accept that there is something wrong in my setup. I will compare the flow main app vs test cases and get back ;)

gimesketvirtadieni:
I just noticed that probably you used MAC for this test (could you confirm your platform?), whereas I use Linux. For my app I use g++ 5.2.1 whereas for tests scripts use g++ 4.8

I will explore if this is related.

KjellKod:
Same on both. 

screen shot 2016-07-11 at 2 13 24 pm


screen shot 2016-07-11 at 2 14 04 pm

The diff is only this, there is no change in the logworker in my test:
sinkhandle.hpp
screen shot 2016-07-11 at 2 14 51 pm

KjellKod:
The test diff is this: 

screen shot 2016-07-11 at 2 19 53 pm

gimesketvirtadieni:
I just installed fresh Kubuntu 16.04 32 bit VM. Then downloaded g3log master, made 2 changes to the code (one was sink() method and one was print out use_count), configured your travis script to use g++-5 and run unit tests... Guess what, I got use_count=3! Some woodoo magic :(

I can see both you and travis (I tested travis and got use_count=2) use 64 bit architecture, whereas I use 32-bit arch. Tomorrow I will test on fresh 64 bits VM and let you know.

test_unit/test_sink.cpp
```diff
@@ -40,6 +40,28 @@ using namespace g3;
...
+   {
+      auto worker = g3::LogWorker::createLogWorker();
+      auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
+      EXPECT_FALSE(flag->load());
+      EXPECT_TRUE(0 == count->load());
+
+      LogMessagePtr message{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
+      message.get()->write().append("this message should trigger an atomic increment at the sink");
+      worker->save(message);
+      worker->removeSink(std::move(handle));
```
KjellKod: You have a race condition here.

Also. To follow your reasoning in this PR. You should verify that sending after the sink is removed should not increment the count.

I'm still missing a test that verifies that the sink is actually destroyed.
To quell the fears that the sink is kept alive I think you need to add another test for that

KjellKod: Regarding the race. Don't feel that you have to follow too closely the test above. It uses RAII to enforce the flush. For your test I think you need to rely on something else.

Example:
1. For the first message you can "cheat" by saving directly to the sink through it's handle
2. Then wait for the returned future, and verify increment of count
3. After this you can remove the sink
4. Then sending a message exactly as you are already doing. No sink should then exist to receive that message
5. Enforce flush of logs through RAII
5. Verify that the last log entry did not increment the count (since the sink was gone).

Doing it this way will remove the race between sending and checking for he counter increment

KjellKod: Or if you come up with easier/smarter way that should work too ;)

KjellKod:
@gimesketvirtadieni regarding the Ubuntu testing: I would update the unit tests first to be race free, as well as the missing destruction test. Let's use that to verify, if possible the lambda issues and implementation

test_unit/test_sink.cpp
gimesketvirtadieni: Could you explain scenario for race condition you had in mind?

test_unit/test_sink.cpp
```diff
@@ -40,6 +40,28 @@ using namespace g3;
...
+      EXPECT_TRUE(flag->load());
+      EXPECT_TRUE(1 == count->load());
+
+      // TODO: saving message after sink was removed will cause seg fault in worker destructor
```
KjellKod: This comment is bothering me. It should be perfectly fine to write log messages when the sink is removed.
What is causing the seg fault?

test_unit/test_sink.cpp
KjellKod: There is a race between saving the message and checking the flag on line 57.
I've outlined above how you can do this race free.

Please remember that you are dealing with three threads here
Your main thread in the test code
The background worker thread 
The sink thread

The test is written as if it is synchronous but it is not. Hence the race.

KjellKod: Or hang on. You do a "wait" in the removeSink, right? Which essentially creates a synchronization point ... Ergo it is not a race. Sorry for the confusion.

It's still missing the real test logic though. How is this test verifying any of the logic in removing a sink?

test_unit/test_sink.cpp
gimesketvirtadieni: This is what I am trying to figure out. If I understand correctly seg fault happens not when message is posted after sink was removed, but in worker's destructor.
I will let you know how it goes ;)

gimesketvirtadieni:
I figured out why there was a seg fault during issuing a message after removing sink: the problem was with my unit test - I reused the same message object whereas it is "gone" after the first invocation of save method. On the other hand LogWorkerImpl::bgSave produces a seg fault when an empty message is provided and no sinks are registered. Check out my last commit ;)
One more update: regarding amount of references to a sink after sink was added - the problem is not related to the platform or 32/64 bits, this is a timing problem and amount of references depend of how "fast" machine is. I am still looking into this.

KjellKod:
> LogWorkerImpl::bgSave produces a seg fault when an empty message is provided and no sinks are registered. 

What do you mean with an empty message? A unique_ptr containing nullptr? If so, yes. That's a scenario that cannot happen in a normal use case.

Sending a normal message when there are no sinks is totally safe.

test_unit/test_sink.cpp
```diff
@@ -40,6 +40,30 @@ using namespace g3;
...
+      auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
+      EXPECT_FALSE(flag->load());
+      EXPECT_TRUE(0 == count->load());
+
+      LogMessagePtr message1{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
+      message1.get()->write().append("this message should trigger an atomic increment at the sink");
+      worker->save(message1);
+
+      worker->removeSink(std::move(handle));
+      EXPECT_TRUE(flag->load());
```
KjellKod: Yup. According the implementation of the ```ScopedSetTrue``` in `testing_helpers.h`.  This means that the `ScopedSetTrue` sink was removed and destroyed.  So that works for me for verifying that the sink is actually removed.

src/logworker.cpp
```diff
@@ -120,6 +122,26 @@ namespace g3 {
       return std::unique_ptr<LogWorker>(new LogWorker);
    }
 
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
+      auto bg_removesink_call = [this, sink] {
+         _impl._sinks.erase(
+            std::remove(
```
KjellKod: formatting nitpicking: can you put the `std::remove` args on the same line as the `std::remove`?

gimesketvirtadieni:
I think I understand why there was a difference with amount of references to a sink after add / remove operation...

There are five instances of sink ptr involved while adding / removing sinks:
- The first sink ptr is created in worker addSink / removeSink
- One more sink ptr is created in addWrappedSink / removeWrappedSink because it is called passing parameters "by value"
- One more sink ptr is created by lambda capturing it "by value"
- One more sink ptr is created while invoking spawn_task because a lambda copy is created during that call
- One more sink ptr is created and stored in vector of sinks

The next piece of the puzzle is spawn_task and packaged_task functionality. As far as I understand completing the packaged pack is done in two steps:
- Setting promise as completed
- Releasing the context of the packaged task

These two steps are not atomic and in my case client thread gets back control after promise is completed but before packaged_task context is released hence instance number 4 of sink ptr is still valid. It remains valid even after all other instances are deleted when removeSink method is done (this is because instance number 4 of sink ptr lives on different thread) preventing from ensuring that sink's destructor is called inside removeSink method.

One way to resolve this issue is to use references for capturing sink ptr in lambdas: it is safe because of synchronization done using promise / futures.

Check out my last commit, I also removed empty lambda in removeSink because there is no need to clear anything in case using references by add / remove lambdas.

KjellKod:
@gimesketvirtadieni  adding a sink waits until it is finished. The calling thread cannot continue until the sink is finished / added. 

You follow the same philosophy for when removing a sink

screen shot 2016-07-18 at 2 40 57 pm
screen shot 2016-07-18 at 2 38 58 pm

I.e. from the calling thread's perspective it is a synchronized operation. 

> 1. The first sink ptr is created in worker addSink / removeSink
> 2. One more sink ptr is created in addWrappedSink / removeWrappedSink because it is called passing parameters "by value"
> 3. One more sink ptr is created by lambda capturing it "by value"
> 4. One more sink ptr is created while invoking spawn_task because a lambda copy is created during that call
> 5. One more sink ptr is created and stored in vector of sinks

After the client side has called `addSink(...)` or `removeSink(...)`  and the function scope for those has exited all the temporary instances of the `shared_ptr` drops out of scope and the reference count goes down.  From the client side it is `synchronized` so I do not see how this can be an issue unless there's a bug in the implementation ---> but I can of course be wrong. 

Please provide a code snippet that proves otherwise and I'll run that across a number of systems and compilers.

src/logworker.cpp
```diff
@@ -111,7 +113,7 @@ namespace g3 {
    }
 
    void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
+      auto bg_addsink_call = [this, &sink] {_impl._sinks.push_back(sink);};
```
KjellKod: I see what you are doing here but I strongly disagree with it.  Please provide a valid reason that can be proven.  The lambda will not and cannot according to C++ spec, as far as I understand it, stay alive past this function call.

src/logworker.cpp
```diff
@@ -120,6 +122,14 @@ namespace g3 {
       return std::unique_ptr<LogWorker>(new LogWorker);
    }
 
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
+      auto bg_removesink_call = [this, &sink] {
```
KjellKod: pass by reference should never be needed. Adding and removing sinks is not high performant code so I am still waiting for proof that this is an issue. 

If it only surfaces on some type of systems then please provide that information and I will look into it.

gimesketvirtadieni: Naturally my findings sound weird to you :)

I made more testing and I think I can support my theory with some evidences, but to do so I am going to need your help ;)

Please follow the steps below and you will understand this issue:

1) You can take the master copy of g3log. As previously, please add sink() method and print-out of use_count in the first unit test in test_sink.cpp.
For your convinience please comment out all other unit tests, just leave this one. Once compiled and run, you should get following results (nothing new so far):

[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from Sink
[ RUN      ] Sink.OneSink
handle->sink().use_count()=2
[       OK ] Sink.OneSink (505 ms)
[----------] 1 test from Sink (505 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (505 ms total)
[  PASSED  ] 1 test.

2) The next step is to add extra logging and a delay to the standart packaged_task implementation.
Please open standrad future header file (in my case it is located in /usr/include/c++/5/future, in your case it might be different).
You need to find operator () of the packaged_task template and add few lines so it looks like following:
      // Execution
      void
      operator()(_ArgTypes... __args)
      {
	__future_base::_State_base::_S_check(_M_state);
	_M_state->_M_run(std::forward<_ArgTypes>(__args)...);
// 4 lines added below
std::cout << "Before pause in packaged_task" << std::endl;
std::chrono::milliseconds wait{200};
std::this_thread::sleep_for(wait);
std::cout << "After pause in packaged_task" << std::endl;
      }

Naturally you will need to add two extra headers at the top of the file along with the rest of the headers definition:
#include iostream
#include chrono

3) Recompile your test and g3log; I had to remove content of build_travis directory to make sure I have a clean recompile.

4) After clean recompile, run test_sink and tell me what you see. If my theory is correct you should see the following print out:
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from Sink
[ RUN      ] Sink.OneSink
Before pause in packaged_task
handle->sink().use_count()=3
After pause in packaged_task
Before pause in packaged_task
After pause in packaged_task
[       OK ] Sink.OneSink (505 ms)
[----------] 1 test from Sink (505 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (505 ms total)
[  PASSED  ] 1 test.

You can play a bit with pause value.

Basically it proves that if delay between setting the promise and removing context of the packaged_task is big enough (which is the case in my situation) then a valid reference to the sink is left after add / remove sink returns controls to the caller.
That is why current synchronisation based on promise / future is not sufficient in this case.
Please note that nothing is wrong with standat library, because promise / future ensure that only result from other thread is available after wait call, whereas current code relies on an assumption that not just result is available but also result producer's context is destroyed.
There are 3 options (all of them work for me) to solve it:
-> Use weak_ptr as captured value in add / remove lambdas (this is the least performant code)
-> Use empty lambdas to clear the context on the worker thread
-> Capture value by reference in add / remove lambdas (the most performant option)

Was it convincing enough for you? ;)

KjellKod: Thanks @gimesketvirtadieni yes. I could see the same thing. Interesting. It's almost as if the `packaged_task` should contain the `join()` function too since otherwise you are not guaranteed resource state. 

I don't think this changes anything though as far as the code goes.  Let's go through the alternative, bottom-up
1) pass the reference --> The problem, as you already pointed out is that the lambda will be kept alive for a longer time. By passing a reference to a `std::shared_ptr` you will in these instances end up with a `dangling reference`. Although, likely, completely safe this is opening up a can of worms and goes against the usage philosophy for `share_ptrs`.     This I therefore consider the worst option. 

2. using `weak_ptr`. That's possible but I do not see a real strong reason for this. The issue seems to be a unit test issue only and one that could easily be worked around.   Once the sink is removed from the internal storage it cannot be used anymore. 

Do you see any issues with a few ms delay in removing the actual sink, apart from unit test wise?  

3. The same reasoning goes for the empty lambda. It is a crutch to fix unit test and to me, does not seem to be adding any value to the production code. 

Unless you can shed some light on why it would impact the production code I suggest you remove the dangling reference code and use plain shared_ptr.   The work around for the unit test is simple enough

```
TEST(Sink, KJELL_OneSinkRemove) {
using namespace g3;
   AtomicBoolPtr flag = make_shared < atomic<bool >> (false);
   AtomicIntPtr count = make_shared < atomic<int >> (0);
   {
      auto worker = g3::LogWorker::createLogWorker();
      auto handle = worker->addSink(std2::make_unique<ScopedSetTrue>(flag, count), &ScopedSetTrue::ReceiveMsg);
      EXPECT_FALSE(flag->load());
      EXPECT_TRUE(0 == count->load());

      LogMessagePtr message1{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
      message1.get()->write().append("this message should trigger an atomic increment at the sink");
      worker->save(message1);
      {
         auto realSink = handle->sink();
         std::cout << "instances: " << realSink.use_count() << std::endl;
      }
      EXPECT_FALSE(flag->load());
      worker->removeSink(std::move(handle));
      size_t kMaxWaitMs = 2000;
      std::chrono::milliseconds wait{200};
      std::chrono::milliseconds maxWait{kMaxWaitMs};
      auto start = std::chrono::steady_clock::now();
      auto stopWatch = [&start]() -> size_t{
         auto end = std::chrono::steady_clock::now();
         std::chrono::duration<double> elapsedMs = end - start;
         return elapsedMs.count();
     };


      while(true) {
         auto elapsedMs = stopWatch();
         if (flag->load() || elapsedMs > kMaxWaitMs) {
           break;
         }
         std::this_thread::sleep_for(wait);
      }
      std::cout << "Max wait was: " << stopWatch() << "ms" << std::endl;
      EXPECT_TRUE(flag->load());
      EXPECT_TRUE(1 == count->load());

      LogMessagePtr message2{std2::make_unique<LogMessage>("test", 0, "test", DEBUG)};
      message2.get()->write().append("this message is issued after all sinks are removed");
      worker->save(message2);
   }
   EXPECT_TRUE(flag->load());
   EXPECT_TRUE(1 == count->load());
}

```

gimesketvirtadieni:
I see your point, I discovered this issue while creating unit test, so no wonder these options do seem like a "work-around" :)

To my mind the main factor is whether you want to make add / remove functionality to be sync or async. If you decide to make this functionality sync then remove function must provide guarantee that sink was actually destroyed before call is returned. This may look excessive but think about following scenario:

- a sink is added (it does not matter whether it is sync or async symantic used)
- after some time sink is removed using removeSink method call (let's say async semantic is used)
- after removeSink call returned, the clean-up code that follows next (for example socket close) is not reliable because sink may be still "alive"

If you ask my opinion, I would say you need to consider:
- either use sync semantic for add / remove functionality by using empty lambdas (flying lambdas is not a problem because they will be cleaned up before joining worker thread)
- or provide flush functionality so users can reach sync semantic with async API

Anyway it is your decision ;)

src/logworker.cpp
KjellKod: @gimesketvirtadieni 

For adding a sink this is not something I worry about at all. 

For removing a sink I understand your point in regard to the example with resource cleanup. Also there is no guarantee that the user hasn't  kept their own shared_ptr of the sink. 

I agree with you that after removing the sink it makes sense to provide a **guarantee** that the sink is done processing all enqueued messages.  That way it does not matter whether or not the `future/packaged_task` keeps it alive for some ms or X time units longer. There should be no more processing of the queue after it is `removed` - even if the actual destruction might be somewhat delayed. 

That's a fairly strong guarantee and I think it would satisfy most users. 


It's an excellent point you bring up to provide a `flush()` function or `synchronization()`/'join()' function. I see no problem adding that. 

If you want that then my suggestion would be to add it to 
https://github.com/KjellKod/g3log/blob/master/src/g3log/sink.hpp

I.e. through it you can use the spawn_task and call `wait()` on an empty lambda.   If for sake of argument we call it 'flush()` then the remove sink could look as follows

```
   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
      auto bg_removesink_call = [this, sink] {
         _impl._sinks.erase(std::remove(_impl._sinks.begin(), _impl._sinks.end(), sink), _impl._sinks.end());
      };
      auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
      token_done.wait(); // the sink is now guaranteed to not receive any new messages.
      sink->flush(); // the sink is now guaranteed to have processed all messages that were enqueued.
   }
```

gimesketvirtadieni:
OK, so let's say both add and remove use async semantic (I keep mentioning both operations for the sake of API consistency).

Then I would say that wait for packaged_task to complete is not need neither in add nor in remove methods. Indeed if one issues a message after any of these method calls but before sink is actually added / removed then this message will stay in a queue and will be processed after previous operation completes.

Also to provide syncing functionality both worker and sink APIs should be extended with flush methods. These methods will provide required guarantees:
- sink flush will ensure that all pending messages are processed by a particular sink
- worker flush will ensure that all pending operations (save, add, remove, ...) are processed and processing context is destroyed

In practice these are really small changes providing tons of flexibility for users ;)

Check out my last commit that highlights these ideas.

src/g3log/logworker.hpp
```diff
@@ -109,6 +109,13 @@ namespace g3 {
          removeWrappedSink(sink_handle->sink());
       }
 
+      void flush(bool propogate_to_sinks = true) {
+         g3::spawn_task([propogate_to_sinks] {
+            if (propogate_to_sinks) {
+               // TODO: ...
```
KjellKod: Is this a placeholder for whether or not the flush should be triggering on all sinks as well?

src/logworker.cpp
```diff
@@ -113,21 +113,22 @@ namespace g3 {
    }
 
    void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_addsink_call = [this, &sink] {_impl._sinks.push_back(sink);};
-      auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
-      token_done.wait();
+      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
+      g3::spawn_task(bg_addsink_call, _impl._bg.get());
```
KjellKod: nice change.  The `wait()` call was never needed.

KjellKod: Ah. I just thought of something.  I remembered why I had the `wait()` here.  It was just to be explicit about what's going on and I did not want to return a `std::future<>` (and at the time I wanted to know when the sink was in place)

So even if you don't do the `wait()` here. The actual result of the code hasn't changed.   The `g3::spawn_task(...)` returns a future. Since you are not capturing the future it falls out of scope and so it has the same effect as calling `wait()`.  

I believe you can just use the `send(...)` function instead if you want to avoid waiting for the `std::future`

src/logworker.cpp
```diff
@@ -113,21 +113,22 @@ namespace g3 {
...
-      auto bg_removesink_call = [this, &sink] {
+      auto bg_removesink_call = [this, sink] {
          _impl._sinks.erase(std::remove(_impl._sinks.begin(), _impl._sinks.end(), sink), _impl._sinks.end());
+
+         // TODO: to be added to the sink API
+         //sink->flush();
       };
-      auto token_done = g3::spawn_task(bg_removesink_call, _impl._bg.get());
-      token_done.wait();
+      g3::spawn_task(bg_removesink_call, _impl._bg.get());
```
KjellKod: Use `send(...)` instead

src/logworker.cpp
```diff
@@ -112,14 +114,23 @@ namespace g3 {
 
    void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
       auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
-      auto token_done = g3::spawn_task(bg_addsink_call, _impl._bg.get());
-      token_done.wait();
+      g3::spawn_task(bg_addsink_call, _impl._bg.get());
```
KjellKod: Maybe I commented on the commit instead?  I'm redoing the comment here. 

This code is the same as calling `token_done.wait()`. `spawn_task(...)` returns a `std::future` if a future isn't captured it falls out of scope and it means the same thing as calling `wait()`. 

So. In that regard I think you should use `send(...)` instead. similar to this:
screen shot 2016-07-20 at 8 53 10 am

As you pointed out. Adding a sink does not need to wait since any messages sent after this call will be in the queue and will be received by the new sink.

src/logworker.cpp
```diff
@@ -112,14 +114,23 @@ namespace g3 {
...
    }
 
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
+      auto bg_removesink_call = [this, sink] {
+         _impl._sinks.erase(std::remove(_impl._sinks.begin(), _impl._sinks.end(), sink), _impl._sinks.end());
+
+         // TODO: to be added to the sink API
+         //sink->flush();
+      };
+      g3::spawn_task(bg_removesink_call, _impl._bg.get());
```
KjellKod: See above regarding using `send(...)` instead. 

That said. I do consider these two functions (add/sink) to be inherently different. Where the `add` gives a sink that is currently empty and the state is known. 

`remove(...)` removes a sink but the sink will continue to process the messages until it's done. 

There's a fine balance between `flexibility` and `robustness`.  I think the `remove` should provide the only guarantees it can provide. 
1.It is removed from the sink list (done here)
2.No more messages are to be processed once it's removed (not yet done). 

I.e. I'm fine with **no** guarantees regarding destruction of the sink. Since the user can grab the sink that guarantee is impossible to give.  However I really think the 2.guarantee needs to be given. 

I.e. last lines should read (pseudo code0
```
send(bg_removesink_call);
sink->flush())
```

KjellKod: With the above changes it is less flexible than you suggested but is more robust and easy to reason about.

src/g3log/logworker.hpp
KjellKod: I prefer explicit APIs to default parameters. 

If you want something to is passed onto the sinks to ensure they are synched also it should instead be split into two functions following the SRP principle. 

pseudo code
```
void flushWorker() {
   auto doNothing = []() {};
   g3::spawn_task(doNothing).wait();
```

```
void flushAllSinks() {
  auto synchronized = []() {
     for (auto& sink : _impl._sinks) {
        sink->flush();
     }
};

spawn_task(synchronized).wait();
}
```

src/logworker.cpp
```diff
@@ -29,6 +29,8 @@ namespace g3 {
 
       if (_sinks.empty()) {
          std::string err_msg {"g3logworker has no sinks. Message: ["};
+
+         // TODO: following line will cause seg fault if empty msgPtr is provided
```
KjellKod: I don't think the TODO is needed. Here is no way for a msg to be nullptr except through unit tests

src/logworker.cpp
```diff
@@ -111,15 +113,38 @@ namespace g3 {
    }
 
    void LogWorker::addWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
-      auto bg_addsink_call = [this, sink] {_impl._sinks.push_back(sink);};
+
+      // capturing sink by reference is safe because this method waits for the completion of this task
+      // passing sink by value may lead to keeping it in other thread's context even after this method returns
```
KjellKod: You just proved that it wasn't safe. That it could be kept alive longer. 

I.e reference to shared_ptr is a not OK

src/logworker.cpp
```diff
@@ -111,15 +113,38 @@ namespace g3 {
...
    std::unique_ptr<LogWorker> LogWorker::createLogWorker() {
       return std::unique_ptr<LogWorker>(new LogWorker);
    }
 
+   void LogWorker::removeWrappedSink(std::shared_ptr<g3::internal::SinkWrapper> sink) {
+
+      // capturing sink by reference is safe because this method waits for this task to complete
+      // passing sink by value may lead to keeping it in other thread's context even after this method returns
+      auto bg_removesink_call = [this, &sink] {
+
```
KjellKod: See above

PR #108, opened 2016-07-30
Warning braces loglevels clang 3.7

jkhoogland:


src/g3log/loglevels.hpp
```diff
@@ -70,12 +70,12 @@ namespace g3 {
 }
 
 #if (defined(CHANGE_G3LOG_DEBUG_TO_DBUG))
-const LEVELS DBUG {g3::kDebugVaulue, {"DEBUG"}},
+const LEVELS DBUG {g3::kDebugVaulue, "DEBUG"},
```
KjellKod: My suggestion is to keep the braces and instead changing the constructor to take a std::string.
Ref the discussion regarding this issue

KjellKod: The warning should go away when you do that

KjellKod:
:+1:

PR #124, opened 2016-10-04
wide character support tested on windows

AndreasSchoenle:


src/g3log/logcapture.hpp
```diff
@@ -57,10 +57,20 @@ struct LogCapture {
    std::ostringstream &stream() {
       return _stream;
    }
+#ifdef ENABLE_WIN_WSTRING_SUPPPORT
+   void capturef(const wchar_t *printf_like_message, ...) __attribute__((format(wprintf, 2, 3))); // 2,3 ref:  http://www.codemaestro.com/reviews/18
 
 
+																							  /// prettifying API for this completely open struct
+   std::wostringstream &wstream() {
```
KjellKod: see typedef question below

src/g3log/g3log.hpp
```diff
@@ -149,6 +154,14 @@ namespace g3 {
 #define CHECK(boolean_expression)        \
    if (false == (boolean_expression))  INTERNAL_CONTRACT_MESSAGE(#boolean_expression).stream()
 
+#ifdef ENABLE_WIN_WSTRING_SUPPPORT
+#	define LOGW(level) if(g3::logLevel(level)) INTERNAL_LOG_MESSAGE(level).wstream()
```
KjellKod: 1. Why would you define a different `LOGW` API if you anyways go through the wide char support in the LogMessage?

2. What does it give you to have access to both, when the `saveMessage` is changed?

src/g3log/logmessage.hpp
```diff
@@ -92,9 +98,9 @@ namespace g3 {
       LEVELS _level;
       std::string _expression; // only with content for CHECK(...) calls
       mutable std::string _message;
-
-
-
+#ifdef ENABLE_WIN_WSTRING_SUPPPORT
```
AndreasSchoenle: The reason is that even in a windows program many log messages can and will be prepared as narrow strings. Dropping narrow string support would dramatically decrease performance in these cases as a utf8 -> utf16 conversion would have to be done before or during the logging request (i.e. in the execution thread).
Vice versa, the actual reason why anyone would want Utf-16 support in g3log is that converting log strings that are more quickly prepared in Utf-16 (anything involving strings coming from or going to the windows API) to Utf-8 before logging them will take its toll on the execution thread. 
Thus, from my point of view moving g2log to wchar_t only does not gain us anything. In this case it would be better to just prepare a layer that converts logging request from Utf-16 to Utf-8 in a convenient way.

My Idea was, on the other hand, to do all conversions in the logging thread and allowing sinks to deal with things like encoding and conversion.

src/g3log/logcapture.hpp
AndreasSchoenle: see answer above

src/g3log/g3log.hpp
AndreasSchoenle: The gain is only in performance. All is based on the assumption that at this point I do not know which encoding the sink will prefer and that I want to log as quickly as possible. Thus when having the choice I would always send an Utf-8 encoded narrow string. However, when I want to log a windows error message or some string received from the GUI which is in Utf-16 I would log it as a std::wstring and let the sink worry about it.

src/g3log/g3log.hpp
```diff
@@ -94,8 +94,13 @@ namespace g3 {
       bool isLoggingInitialized();
 
       // Save the created LogMessage to any existing sinks
+#ifdef ENABLE_WIN_WSTRING_SUPPPORT
+	  void saveMessage(const char *message, const wchar_t *wmessage, const char *file, int line, const char *function, const LEVELS &level,
```
AndreasSchoenle: I understand and I had thought about it a long time. Code duplication was one issue, the other that if you decide to keep the concept of conversion in the sink, it is absolutely reasonable to add the stack trace and fatal message stuff as Utf-8 narrow strings (as most or all of it is plain ascii anyhow).

AndreasSchoenle:
I guess it all boils down to deciding between

1. No Utf-16 supoort at all
2. Keep both std::string and/or std::wstring in the execution thread, convert in the log thread
3. Provide Utf-16 only version through typedef and/or templates

As outlined above, I see no real use in option 3. Most messages will ultimately be converted to Utf-8 encoding before writing them to a file in most sane environments. 

Option 1 as absolutely valid, it will sacrifice some performance in windows environment but the user can add their favorite conversion tool. For me it would merely mean to write
LOG(DBUG) << Utf8(L"Wstring message");
instead of 
LOGW(DBUG) << L"Wstring message"
and I can probably absolutely live with it. 

I did prefer option 2 because the philosophy in your library seemed to be to avoid as much work as possible in the execution thread and because it allows the user to more or less transparently pass both std::string and std::wstring messages through the infrastructure without the framework even having to know the actual encoding (when using custom sinks).
I agree the solution in my merge request lacks a little in the elegance department. (LOG and LOGW APIs etc.) but it was the best I could come up with in the time I had that allows this and does not require major code additions or changes that could introduce issues. 

If you like option 2 and want to consider it I can look into the saveMessage thing, of course.

KjellKod:
I'm fine with adding it but I would like to untangle this.

My suggestion would be to keep the LOGW and CHECKW as you have today but removing the ifdefs.
If people want wide character support they can use those macros and if they don't it'll stay the same.

That means you need to split some functions and making a smart choice which of the "saveMessage" function to call.

There should be no performance penalty for not using wide characters

KjellKod:
Any continued progress on this?

AndreasSchoenle:
Sorry, I have been quite busy but I definitely plan to continue on this. 

One question: Would it be ok to have the std::wstring member present always (it would be a slight performance hit, as the empty object would be moved and copied) and the "new" signature of saveMessage() including the const wchar_t * parameter? 

Also there are two or three smaller pitfalls that I wanted to address before continuing with it: A compiler error on ARM (don't really  understand why it happens but it cannot deal with SinkWrapper being pure virtual (the problem was solved when defining the send() method, most probably the ARM (I am stuck with gcc 4.7) compiler tries to instantiate somewhere where it is not necessary) and a few CMake related things. 

I wanted to also add them to the issue list and check back with you about them. I have little hope of making time for it before Christmas but let me know your thoughts about the wchar_t issue and I will try to squeeze it in at some point and deal with the other issues later.

KjellKod:
It's fine with wstring to be present

AndreasSchoenle:
I have changed it such that it is always added. The overhead (runtime) is creating and then moving around an empty std::wstring. Codewise it adds a second version of capturef(). This could be avoided by making LogCapture a template class - but as discussed above, this comes with quite some re-coding in other areas and it seems like a small price to pay for clarity.
Unfortunately gcc 4.7 (in fact below 5) have no codecvt support, so conversion from wstring to utf-8 std::string cannot be accomplished without pulling in e.g. boost::locale. This concerns only the message() function, i.e. the default formatting of a message AND if LOGW was used AND gcc < 5.0, so it seemed to me like not good reason to add a dependency or write huge amounts of code.

CMakeLists.txt
```diff
@@ -48,6 +48,10 @@
 
 cmake_minimum_required (VERSION 3.1)
 ENABLE_LANGUAGE(CXX)
+if( POLICY CMP0054 )
```
KjellKod: can be removed

src/g3log.cpp
```diff
@@ -153,17 +153,16 @@ namespace g3 {
...
+         const char *boolean_expression, int fatal_signal, const char *stack_trace) {
 
-      /** explicits copy of all input. This is makes it possibly to use g3log across dynamically loaded libraries
-      * i.e. (dlopen + dlsym)  */
-      void saveMessage(const char *entry, const char *file, int line, const char *function, const LEVELS &level,
-                       const char *boolean_expression, int fatal_signal, const char *stack_trace) {
          LEVELS msgLevel {level};
          LogMessagePtr message {std2::make_unique<LogMessage>(file, line, function, msgLevel)};
+
+         message.get()->wwrite().append(wentry);
```
KjellKod: so if there is no data here, this `wentry` is nullptr, right?  Can we make that check before writing, it's either one or the other that is ```nullptr```, correct?

KjellKod: The reason why I'm mentioning this is that it might be a platform dependent issue to write ```nullptr```

src/logmessage.cpp
```diff
@@ -108,7 +115,19 @@ namespace g3 {
...
    }
 
-
+   std::string LogMessage::message() const {
+      std::string msg;
+      if (!_wmessage.empty())
+      {
+#if !defined(__GNUC__) || (__GNUC__ >= 5)
+         std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> convert;
+         msg.append(_message.append(convert.to_bytes(_wmessage)));
```
KjellKod: So this means that although we have LOGW etc the ```message```` function is only implemented for the ```std::wstring``` for non-GNU. If for some reason someone want to use LOGW on a GNUC it essentially will render the ```LOGW``` call meaningless, right?

src/time.cpp
```diff
@@ -100,11 +100,10 @@ namespace g3 {
       ts->tv_nsec = now.tv_usec * 1000;
       return 0;
 #elif(defined(WIN32) || defined(_WIN32) || defined(__WIN32__))
```
KjellKod: please note that this will clash with your time pull request. Once that is merged in, which should be soon, please remember to update this PR.

src/g3log/g3log.hpp
```diff
@@ -136,18 +136,23 @@ namespace g3 {
 
 // LOG(level) is the API for the stream log
 #define LOG(level) if(g3::logLevel(level)) INTERNAL_LOG_MESSAGE(level).stream()
-
+#define LOGW(level) if(g3::logLevel(level)) INTERNAL_LOG_MESSAGE(level).wstream()
```
KjellKod: They also need to be documented in the API.markdown.

src/logmessage.cpp
AndreasSchoenle: Yes, at least gcc 4.7 does not know the codecvt header and has no wstring_convert. Current gcc does, though: [ideone example](http://ideone.com/ft0w9k)
I am not sure which stdlibc++ is needed for this to work. Here I just assume gcc 5 or above is ok. The proper way to do this is to move decision to the cmake file and include a compilation test and then have a macro WSTRING_CONVERT_AVAILABLE that decides here. 
Should I do it that way?

Also I should mention that what I do here is EXTREMELY inefficient. The bulk of the time is spent constructing the convert object. This is meant as a fall-back when no custom sink is used. wstring_convert [is not thread safe](http://stackoverflow.com/questions/16644257/are-member-functions-of-stdwstring-convert-thread-safe), so using a static wstring_convert would entail synchronization or the guarantee that the message() function is only ever called from one thread or a thread local wstring_convert object. 
What is your feeling? How important is performance on the log worker side? I would assume that whoever uses wstring logging has a custom sink.

KjellKod: I think we can skip the compilation test but have a CMAKE generated `#define` (please see how the options are handled). 

1. As long as it is the sink that deals with the cpu load I think it is fine

2. Why not splitting out the functionality into a ```wmessage()``` function. It could be something like
```
bool hasWMessage(){ ... }
std::string wmessage() { .... }
```
That way a sink can easily check if the message struct contains a `wstring` and run the ```wmessage()``` function which would then contain the `convert` functionality you currently have in the ```std::string message() {...}``` function. 

3. If you make the convert object a ```std::thread_local``` then it will be thread safe and it will only be instantiated once.

KjellKod: very nice work by the way @AndreasSchoenle

KjellKod:
@AndreasSchoenle please see the conflict and the PR that cause it ;)  https://github.com/KjellKod/g3log/pull/137

KjellKod:
@AndreasSchoenle : If you are too busy then I can try to take over this PR, to address the conflicts and the remaining comments?

AndreasSchoenle:
@KjellKod sorry that I did not get around to do it. I was indeed kind of dug-in but may be able to make some time next week to look into it. I remember that I also wanted to look into the cmake generated defines (to have the file put into the build directory, not the source dir and to make sure it is installed along with the other headers). 
I will get back to you if it becomes clear that I cannot address the remaining issues within next week?

KjellKod:
Sounds good.

If you can't take it then I can always try to take care of it myself.  Let's sync if it comes to that.

The build defines are installed now with the other headers. 

Cheers 
Kjell

AndreasSchoenle:
I looked into it and found that master had moved too much to simply patch my merge request. I will try to move to 1.3 in my production code  but it seems like a few windows-specific things need to be fixed before I can start.
The plan is now to get 1.3 ready (see my latest pull request) and start from master re-adding the wide-char support keeping all your comments and suggestions in mind. 
We could thus close the current pull request along with cmake_changes unless you think it is a bad plan.

KjellKod:
I agree with the plan. It's the approach I would have taken as well

Issue #125, opened 2016-10-17
CMake warnings because of unclear policy version ("CMP0054") in 'Build.cmake'.


dasmysh:
A current version of CMake will show several warnings because it is not clear which policy to use when comparing string literals (see [CMake Help](https://cmake.org/cmake/help/v3.1/policy/CMP0054.html))

The issue can be resolved by putting
`cmake_policy(SET CMP0054 NEW)`
at the beginning of the file.

KjellKod:
Perfect. Can you put up a pull request?

dasmysh:
Did that. See #126 .

KjellKod:
I'll see if I can update Travis. The CMake changes is currently braking the CI for this pull request

KjellKod:
bump

KjellKod:
merged: https://github.com/KjellKod/g3log/pull/136

PR #126, opened 2016-10-17
Updated cmake version to correct string compare behaviour.

dasmysh:
When building g3log with newer CMake versions (>3.1) there will be several warnings about a change in string comparison behavior/policy (see #125 for more info). The old policy can lead to some unexpected comparison results.

To be able to use the new policy I had to require a newer CMake version (at least I think that is the most elegant way). I am not sure if there was any reason for using an older version but if not it might be a good idea to update.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,7 @@
 #
 # ============================================================================
 
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 3.1)
 ENABLE_LANGUAGE(CXX)
 set(CMAKE_BUILD_TYPE Release)
 


KjellKod:
I have (or you) to update Travis to make this work. Currently this breaks our CI.

CMakeLists.txt
```diff
@@ -46,7 +46,7 @@
 #
 # ============================================================================
 
-cmake_minimum_required (VERSION 2.8)
```
KjellKod: This was on the backburner for some time but I've updated travis to use a newer CMake version. 

In the issue you commented on that the following was needed to remove the possibly weird behaviour
```
cmake_policy(SET CMP0054 NEW)
```

Shouldn't that be part of this pull request too?

KjellKod:
Ref: https://github.com/KjellKod/g3log/pull/135

CMakeLists.txt
dasmysh: By setting the minimum required version the policy is implicitly set, so it is not needed in the pull request.

PR #128, opened 2016-10-19
Fix compilation errors when CHANGE_G3LOG_DEBUG_TO_DBUG is enabled-

AndreasSchoenle:
Fixing the following issue:
When using DBUG instead of DEBUG level through cmake unit tests and main programs would no longer compile.


KjellKod:
👍

Issue #15, opened 2015-06-24
capturef should use the _Printf_format_string_ annotation on MSVC


saul:
See https://msdn.microsoft.com/en-us/library/vstudio/ms235402(v=vs.100).aspx

The capturef signature can be changed to:
```cpp
void capturef(_Printf_format_string_ const char* printf_like_message, ...) __attribute__((format(printf, 2, 3)));
```

It may make more sense to have two seperate declarations - one for GCC, one for MSVC and one for others.

KjellKod:
Hi Saul, 

It sounds like a good suggestion.  If you want to expediate this, feel free to do a pull request. 
Cheers
Kjell

KjellKod:
@saul this issue is nearing one year mark. So far no one else has chimed in and I have not seen a pull request to deal with the Windows part of this so it is a candidate for being closed.

KjellKod:
resolved here: https://github.com/KjellKod/g3log/pull/129

PR #129, opened 2016-10-19
MSVC printf format specifier

AndreasSchoenle:
Proposal to address #15.


KjellKod:
👍  thanks for making sure it didn't break Linux and OSX

PR #135, opened 2016-12-04
Cmake changes

AndreasSchoenle:
These changes are tested with CMake 3.5.2 but I tested with cmake_policy(VERSION 2.8), so it should work for the minimum required version. However, I think it would be useful to update the minimum required version of CMake as a lot of useful stuff has been introduced since then. BTW using at least 3.1 will make the policy CMP0054 if-clause obsolete.

Here is what this patch does:

- Provide an INSTALL target also for MSVC
- Use message(STATUS "...") wherever it is not an error or warning to allow spotting of errors in CMake output easier.
- Allow a possible toolchain file to set  ${TOOLCHAIN_COMPILER_FLAGS} in CMake in order to introduce toolchain specific (additional) compiler flags. If this variable is not set, nothing changes.
- In WIN32, change the name of the static library to g3logger_static when the DLL is also built. On WIN32 the DLL should be the default (**I would even suggest removing the static library alltogether or at least remove the option for adding the DLL and add it always**), as usually one will use g3log across multiple DLLs in one project and ensure that every DLL uses the same logger (g_logger_instance).
Not sure, BTW, about the behavior on Linux if several shared libraries use the static lib and are part of one executable. Maybe the situation is similar and the static library should be removed all together?
- Moved the generated header file out of the source tree to enable building for several platforms with different settings out of the same source without the need to re-compile (or: ensure that cmake re-runs when changing platforms). The header is installed with the rest to the /include subdir. The include paths for the compiler are adjusted to find the generated header. Technically the include statement should probably be changed to use <> instead of "".

I did not test this on linux very well as I only compile on Windows and for Win32, Win64 and my very vendor specific ARM platform but I tried to make sure that nothing changes on Linux with this patch.


KjellKod:

CMakeLists.txt
```diff
@@ -48,6 +48,9 @@
 
 cmake_minimum_required (VERSION 3.1)
 ENABLE_LANGUAGE(CXX)
+if( POLICY CMP0054 )
```
KjellKod: I don't think this is needed with cmake 3.1, right?

CMakeLists.txt
```diff
@@ -109,8 +129,22 @@ endif()
...
 
 
+if( WIN32 AND G3LOG_SHARED_LIBRARY )
+   install( PROGRAMS $<TARGET_PDB_FILE:g3logger_shared>
+		DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+		COMPONENT libraries
+	)
+endif()
 
+INSTALL( TARGETS g3logger ${G3LOG_SHARED_LIBRARY}
```
KjellKod: please explain the rationale between moving the install instructions from the cpack to the cmake file

CMakeLists.txt
AndreasSchoenle: Correct, the warning goes away starting with 3.1

AndreasSchoenle:
We usually put the install directive next to where the target is created (in fact we usually create our targets with custom scripts that take care of all that).
Thus when you add new targets in subdirectories or tie some subdirs into another cmake project as externals the master files do not need to be edited. It seems common practice and makes sense to me.

In this particular case I could leave the rest of the CPack file untouched and make sure that windows sees the install directive.

Build.cmake
```diff
@@ -92,10 +92,17 @@ ENDIF()
...
    #MESSAGE("  g3logger files: [${SRC_FILES}]")
    add_library(g3logger ${SRC_FILES})
-   set_target_properties(g3logger PROPERTIES
-      LINKER_LANGUAGE CXX
-      OUTPUT_NAME g3logger
-      CLEAN_DIRECT_OUTPUT 1)
+   if (WIN32 AND ADD_BUILD_WIN_SHARED)
+		set_target_properties(g3logger PROPERTIES
+		  LINKER_LANGUAGE CXX
+		  OUTPUT_NAME g3logger_static
```
AndreasSchoenle: will do.

CMakeLists.txt
AndreasSchoenle: sorry. I left the answer below.

KjellKod:
@AndreasSchoenle  please see the conflict and the PR that caused it: https://github.com/KjellKod/g3log/pull/145

AndreasSchoenle:
Became obsolete, as shared library is now the default.

KjellKod:
Thanks 

Please see the raised issues however.
There were people using the static line so I'll bring the static build option back again

AndreasSchoenle:
OK. Not sure what you meant by raised issues? Static as an option is quite ok in my opinion. Would you introduce it as an option?
See comment in other merge request.

KjellKod:
@AndreasSchoenle  would you mind opening up a new pull request that only adds the windows install target as that is currently missing?

PR #137, opened 2016-12-04
fix timing inaccuracy on windows

AndreasSchoenle:
The timespec_get function in windows gives only file time precision. This implementation gives ns precision on my Win7 by using the high_resolution_clock which resolves to steady_clock at least in my version of MSVC. 

The 64 bit integer will overflow in 250 years (2^63/1e9/3600/24/365=292.5) but storing the offset in a timespec struct would bloat the function and the math a bit. I was torn ...


KjellKod:

src/time.cpp
```diff
@@ -87,20 +87,27 @@ namespace g3 {
...
-      struct timeval now = {};
-      int rv = gettimeofday(&now, nullptr);
-      if (-1 == rv) {
-         return rv;
-      }
-      // error mode. just return sec, microsecond
-      ts->tv_sec  = now.tv_sec;
-      ts->tv_nsec = now.tv_usec * 1000;
-      return 0;
+	   // std::timespec_get or posix clock_gettime)(...) are not
```
KjellKod: please adjust formatting (as previously discussed over email)

src/time.cpp
```diff
@@ -87,20 +87,27 @@ namespace g3 {
...
+	   if (-1 == rv) {
+		   return rv;
+	   }
+	   // error mode. just return sec, microsecond
+	   ts->tv_sec = now.tv_sec;
+	   ts->tv_nsec = now.tv_usec * 1000;
+	   return 0;
 #elif(defined(WIN32) || defined(_WIN32) || defined(__WIN32__))
-	   return timespec_get(ts, TIME_UTC);
+	   static auto os =
```
KjellKod: Hmm. I haven't seen this exercise before.  Can you explain what happens here. What exactly is the `os` that you get when you subtract `system clock time since epoch` with` high resolution clock time since epoc`?

src/time.cpp
```diff
@@ -87,20 +87,27 @@ namespace g3 {
...
+		   std::chrono::time_point_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now()).time_since_epoch() -
+		   std::chrono::time_point_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now()).time_since_epoch();
+
+	   auto now_ns = (std::chrono::time_point_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now()).time_since_epoch() + os).count();
+	   ts ->tv_sec = now_ns / 1000000000;
+	   ts ->tv_nsec = now_ns % 1000000000;
```
KjellKod: alright. you have lost me. how is line 109 giving the nanoseconds?
ref: http://ideone.com/FRYjdo

AndreasSchoenle: It calculates the remainder of a division by 1e9 (which in turn yields the full seconds). Most compilers will make a single integer division with remainder out of the two statements (thus I did use the second divison, which % effectively is instead of subtraction tv_sec * 1e9.

src/time.cpp
AndreasSchoenle: the time since epoch for the steady_clock is not necessarily really the time since 1970. It usually is the time since program start. 
Thus, here I calculate the offset between this (different) starting point and the real start of the epoch as reported by the system clock during the first call with the precision of the system clock. That means that in absolute terms time stamps will later have system clock accuracy but relative times will have the precision of the high resolution clock.
For my previous logging solutions had used only the steady clock and had the sink side calculate the absolute time in a similar way (by adding the offset). In fact it is an option to move some of this math to the logging thread (by allowing the message to be modified before passing it to the sinks). This would save time in the execution thread. Similarly splitPath() could be moved there, leaving _file empty in saveMessage and only populating _file_path (in case you accept #138)

src/time.cpp
```diff
@@ -96,15 +96,21 @@ namespace g3 {
...
       }
       // error mode. just return sec, microsecond
-      ts->tv_sec  = now.tv_sec;
+      ts->tv_sec = now.tv_sec;
       ts->tv_nsec = now.tv_usec * 1000;
       return 0;
 #elif(defined(WIN32) || defined(_WIN32) || defined(__WIN32__))
-	   return timespec_get(ts, TIME_UTC);
+      static auto os =
+         std::chrono::time_point_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now()).time_since_epoch() -
```
KjellKod: I think it is good to still keep it within this wrapped ```timespec_get``` function though as it speaks volumes what it does by re-using the name.

KjellKod:
@AndreasSchoenle please see my pull reqest to this pull request :)  https://github.com/AbberiorInstruments/g3log/pull/1

have a great weekend.

AndreasSchoenle:
Thanks for the pull request and sorry I did not get around to any work on this in the past weeks. 
Hopefully I will get some work into the other pull requests, too. 
For now, have a good week and a great Holiday season.

KjellKod:
@AndreasSchoenle great work and lots of thanks for the effort! It is highly appreciated

👍    & Happy Holidays.

PR #138, opened 2016-12-04
Include file path only

AndreasSchoenle:


AndreasSchoenle:
Keep full source file path - this is nice for e.g. Debug output sinks. We can rely on the const char * _file_path  to remain valid (__FILE__ macro). 
Possibly one could even move the splitting of the filename to the sink and save the _file member?

KjellKod:

src/g3log/logmessage.hpp
```diff
@@ -73,7 +76,7 @@ namespace g3 {
       LogMessage& operator=(LogMessage other);
 
 
-      LogMessage(const std::string& file, const int line, const std::string& function, const LEVELS& level);
+      LogMessage(const char * file, const int line, const std::string& function, const LEVELS& level);
```
KjellKod: Unfortunately you cannot rely on the `char*` to remain valid across library borders. That's why the data is copied into the class. One runtime library can delete the data in a different way than expected in the LogMessage receiving runtime.  

It is *probably* safe to change this but I know it works when copied so please keep the copy paradigm. i.e. keeping it as std::string.

AndreasSchoenle: makes sense. will do.

src/g3log/logmessage.hpp
```diff
@@ -87,6 +90,7 @@ namespace g3 {
       timespec _timestamp;
       std::thread::id _call_thread_id;
       std::string _file;
+	  std::string _file_path;
```
KjellKod: please adjust formatting

src/logmessage.cpp
```diff
@@ -123,10 +123,11 @@ namespace g3 {
    }
 
 
-   LogMessage::LogMessage(const std::string& file, const int line,
+   LogMessage::LogMessage(const char * file, const int line,
```
KjellKod: then I think the std::string parameter can be left too as it was originally, right?

AndreasSchoenle: yes, my bad.

KjellKod:
:+1:

PR #139, opened 2016-12-04
Add levels only

AndreasSchoenle:
"Dynamic" adding of log levels at initialization time (without touching actual g3log headers).

### DIFF
--- a/src/g3log/loglevels.hpp
+++ b/src/g3log/loglevels.hpp
@@ -115,7 +115,7 @@ namespace g3 {
    namespace only_change_at_initialization {
       // Enable/Disable a log level {DEBUG,INFO,WARNING,FATAL}
       void setLogLevel(LEVELS level, bool enabled_status);
-	  void addLogLevel(LEVELS log_level);
+	  void addLogLevel(LEVELS level, bool default_enabled_status = true);
       std::string printLevels();
       void reset();
 
--- a/src/g3log/loglevels.hpp
+++ b/src/g3log/loglevels.hpp
@@ -115,9 +115,9 @@ namespace g3 {
    namespace only_change_at_initialization {
       // Enable/Disable a log level {DEBUG,INFO,WARNING,FATAL}
       void setLogLevel(LEVELS level, bool enabled_status);
-	  void addLogLevel(LEVELS level, bool default_enabled_status = true);
+	  void addLogLevel(LEVELS level, bool enabled_status = true);
       std::string printLevels();
-      void reset();
+      void reset(bool remove_custom_levels = false);
 
    } // only_change_at_initialization
 #endif
--- a/src/g3log/loglevels.hpp
+++ b/src/g3log/loglevels.hpp
@@ -115,9 +115,8 @@ namespace g3 {
    namespace only_change_at_initialization {
       // Enable/Disable a log level {DEBUG,INFO,WARNING,FATAL}
       void setLogLevel(LEVELS level, bool enabled_status);
-	  void addLogLevel(LEVELS level, bool enabled_status = true);
       std::string printLevels();
-      void reset(bool remove_custom_levels = false);
+      void reset();
 
    } // only_change_at_initialization
 #endif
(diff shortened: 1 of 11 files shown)


src/loglevels.cpp
```diff
@@ -51,6 +52,11 @@ namespace g3 {
          internal::g_log_level_status[level].get().store(enabled, std::memory_order_release);
       }
 
+	  void addLogLevel(LEVELS log_level) {
+		  internal::g_default_log_level_status[log_level.value].get().store(true, std::memory_order_release);
+		  internal::g_log_level_status[log_level.value].get().store(true, std::memory_order_release);
```
KjellKod: OK, so addLogLevel adds it to the default set as well.

Why not for the second paragraph call the already existing function "SetLogLevel".?

KjellKod:
Looks pretty good. 

Please add unit tests and explain the new function in the API.markdown

src/loglevels.cpp
AndreasSchoenle: You are absolutely right. I will change it with the next commit.

AndreasSchoenle:
I added a description to the API markdown and made some changes that seemed necessary including the reuse of setLogLevel in addLogLevel.
On second thought the added functionality is not all that valuable, though as dynamic log levels are supposed to be changed only at initialization (how strict is this limitation?). After all the only thing this pull request will do is to allow calling reset() without removing custom log levels - something that will be quite seldom in practice as the host program is most probably aware of all available logging levels and their current state.
On the other hand, it has advantages in a test scenario where the host program may not be 'that smart'. I guess what I want to say is that other than the wchar_t stuff this patch is not crucial to my use case.

KjellKod:

src/g3log/loglevels.hpp
```diff
@@ -115,8 +115,9 @@ namespace g3 {
    namespace only_change_at_initialization {
       // Enable/Disable a log level {DEBUG,INFO,WARNING,FATAL}
       void setLogLevel(LEVELS level, bool enabled_status);
+	  void addLogLevel(LEVELS level, bool enabled_status = true);
       std::string printLevels();
-      void reset();
+      void reset(bool remove_custom_levels = false);
```
KjellKod: please remove the boolean. I am a strong believer in SRP and this function has now two responsibilities.

src/loglevels.cpp
```diff
@@ -59,9 +65,10 @@ namespace g3 {
          return levels;
       }
 
-      void reset() {
-         internal::g_log_level_status.clear();
-         internal::g_log_level_status = std::map<int, atomicbool>{{g3::kDebugValue, true}, {INFO.value, true}, {WARNING.value, true}, {FATAL.value, true}};
+      void reset(bool remove_custom_levels) {
+         if (remove_custom_levels)
+            internal::g_default_log_level_status = { { g3::kDebugValue, true },{ INFO.value, true },{ WARNING.value, true },{ FATAL.value, true } };
```
KjellKod: see comment above

KjellKod:
ref

`On second thought the added functionality is not all that valuable, though as dynamic log
 levels are supposed to be changed only at initialization (how strict is this limitation?). 
After all the only thing this pull request will do is to allow calling reset() without removing
 custom log levels - something that will be quite seldom in practice as the host program is
 most probably aware of all available logging levels and their current state.`

Agreed. It is not terribly useful. I was curious as to what your use case was for adding this. I assumed it was something I hadn't encountered before. 

If it isn't something very useful and not (??) a very interesting use case I would suggest closing the pull request but I am still interested in your use case.

AndreasSchoenle:
I had built this because I though I would need it to replace my existing solution. It turns out that I can do it all using existing functionality - not sure any more why I thought I would have to call reset() from a place where I have no knowledge of the intended default levels. 
So, please close it and in case I remember why I thought it was such a good idea, I can still come back to it.

AndreasSchoenle:
I just removed the changes. The only remaining change is now that the unit test does not prematurely end when fatal signal handling is not compiled in.

PR #143, opened 2016-12-21
Couple of improvements

rickyzhang82:
1. Enabled Mac OS X install. I tested it on Mac OS X. It is ok to install /usr/local like Linux

2. Added Support for find_package(G3LOG), use ${G3LOG_INCLUDE_DIRS} for header and ${G3LOG_LIBRARIES} for link library

3. Fix new line issues. The new line should be added to the end. Personally speaking, it is quite ugly put new line in the front of message line.


cmake/G3LOGConfig.cmake
```diff
@@ -0,0 +1,32 @@
...
+#
+# This module will set the following variables in your project:
+#
+# ``G3LOG_INCLUDE_DIRS``
+#   where to find g3log.hpp, etc.
+# ``G3LOG_LIBRARIES``
+#   the libraries to link against to use libg3logger.
+#   that includes libg3logger library files.
+# ``G3LOG_FOUND``
+#   If false, do not try to use G3LOG.
```
KjellKod: nice! thanks

src/logmessage.cpp
```diff
@@ -27,35 +27,35 @@ namespace g3 {
...
       return out;
    }
 
    // helper for fatal signal
    std::string  fatalSignalToString(const LogMessage& msg) {
       std::string out; // clear any previous text and formatting
-      out.append("\n" + msg.timestamp()
+      out.append(msg.timestamp()
                  + "\n\n***** FATAL SIGNAL RECEIVED ******* \n"
-                 + '"' + msg.message() + '"');
```
KjellKod: Please explain reasoning.

cmake/G3LOGConfig.cmake
```diff
@@ -0,0 +1,32 @@
...
+include(FindPackageHandleStandardArgs)
+
+find_path(G3LOG_INCLUDE_DIR g3log/g3log.hpp)
+
+find_library(G3LOG_LIBRARY
+                NAMES libg3logger g3logger)
```
KjellKod: please adjust formatting.

src/filesinkhelper.ipp
```diff
@@ -79,9 +79,7 @@ namespace g3 {
       std::string header() {
          std::ostringstream ss_entry;
          //  Day Month Date Time Year: is written as "%a %b %d %H:%M:%S %Y" and formatted output as : Wed Sep 19 08:28:16 2012
-         ss_entry << "\t\tg3log created log at: " << g3::localtime_formatted(g3::systemtime_now(), "%a %b %d %H:%M:%S %Y") << "\n";
-         ss_entry << "\t\tLOG format: [YYYY/MM/DD hh:mm:ss uuu* LEVEL FILE->FUNCTION:LINE] message";
```
rickyzhang82: No problem, if you insist this. But I will remove \t\t. Please let me know if you agree,

src/logmessage.cpp
rickyzhang82: I get used to Log4j style. If you don't like mine, it is fine.  

1. I don't agree to add new line in beginning of the every message. The new line should be added in the very end of message. It looks very weird to me, when I use sample color cout sink and also my program use std::cout at the same time. My text from std::cout follows g3log message because no new line after g3log message.

2. It is unnecessary to quote the log message.

CMakeLists.txt
```diff
@@ -112,7 +112,7 @@ endif()
 
 
 
-IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+   IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux" OR ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
KjellKod: Thanks for doing this. For future reference it is better to keep pull request according to SRP, i.e. single responsibility principle. here there are multiple targets that you are targeting. Albeit small changes it is still better to split them up.

CPackLists.txt
```diff
@@ -30,6 +30,8 @@
             DESTINATION ${CMAKE_INSTALL_HEADERDIR}/g3log
             COMPONENT headers)
 
+   INSTALL( FILES ${PROJECT_SOURCE_DIR}/cmake/G3LOGConfig.cmake
+            DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/G3LOG)
```
KjellKod: Is this the default install path for cmake on OSX?  It does not seem to be the default install location under Linux: https://cmake.org/Wiki/CMake:How_To_Find_Libraries

src/filesinkhelper.ipp
KjellKod: sounds good

src/logmessage.cpp
```diff
@@ -27,35 +27,35 @@ namespace g3 {
    // helper for setting the normal log details in an entry
    std::string LogDetailsToString(const LogMessage& msg) {
       std::string out;
-      out.append("\n" + msg.timestamp() + "\t"
```
KjellKod: lol. I couldn't care less if the newline is first or last since the end result is the same. I.e this is fine with me. The same goes for the removal of the `"` quotation marks around the log entry. They are not needed of course. 

Since this might affect other users I will call on the to voice their opinion.  They can diff these and voice any objections:
@jkhoogland , @cstamatopoulos , @ctapmex , @AndreasSchoenle 
[original-look.txt](https://github.com/KjellKod/g3log/files/668044/original-look.txt)
[new-look.txt](https://github.com/KjellKod/g3log/files/668045/new-look.txt)

src/logmessage.cpp
KjellKod: FYI: Please also be aware that it is extremely easy to use your own logging sink.
Ref: github.com/KjellKod/g3sinks and the unit tests. Custom formatting can therefore be done very easily.

CMakeLists.txt
KjellKod: No action needed for this PR regarding SRP.

src/logmessage.cpp
rickyzhang82: 1. Yes, I agree it is easy to customize. But the default format should work out of box. For example, the case that user mixed with std::cout and demo ColorCoutSink  doesn't get along well with each other would lead to bad user experiences. That's why I strongly against add new line at the beginning of message.

cmake/G3LOGConfig.cmake
rickyzhang82: No problem.

CMakeLists.txt
rickyzhang82: Sounds like you like CMake PR but not the formatting one. I will separate them later.

CPackLists.txt
rickyzhang82: I confirm that it works in both Mac OS X and Linux. CMake in Mac OSX comes from macport. 
But it really doesn't matter I believe. It is CMake version matters.

See doc : https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html#package-configuration-file

There is a sample below how to create/install package.

src/logmessage.cpp
rickyzhang82: To better demonstrate my motivation, mix std::cout with ColorCoutSink. The new line in the beginning message caused so much trouble. Eg,
``` C++
std::cout<<" Hello"<<endl;
LOG(WARN)<<"Test";
std::cout<<"world"<<endl;
```
The original would look like this

```bash
1.....Hello
2.
3.  LOG: Test world
4. 
```

The new would like 
```bash
1. Hello
2. LOG: Test
3. world
```

CMakeLists.txt
KjellKod: I like them both but prefer to have SRP also for PRs

CPackLists.txt
KjellKod: Yeah I know that it works. Interesting that the CMake guys give conflicting information in regard to this

src/logmessage.cpp
KjellKod: Good point

src/logmessage.cpp
KjellKod: OK

CMakeLists.txt
KjellKod: It's fine as is. Just keep that in mind if there are future PRs coming

src/filesinkhelper.ipp
KjellKod: Great job. If not contrary feedback comes from the called out g3log users I am good with these changes. 


As far as I can see the only thing missing now is tiny formatting issues (CMake) and to put the header back for the beginning of the log file. 


Thanks

rickyzhang82:
See separate PR #144 and #145

PR #144, opened 2016-12-22
Fix message newline issue

rickyzhang82:
Signed-off-by: Ricky Zhang <rickyzhang@gmail.com>

### DIFF
--- a/src/filesinkhelper.ipp
+++ b/src/filesinkhelper.ipp
@@ -79,8 +79,8 @@ namespace g3 {
       std::string header() {
          std::ostringstream ss_entry;
          //  Day Month Date Time Year: is written as "%a %b %d %H:%M:%S %Y" and formatted output as : Wed Sep 19 08:28:16 2012
-         ss_entry << "g3log created log at: " << g3::localtime_formatted(g3::systemtime_now(), "%a %b %d %H:%M:%S %Y") << "\n";
-         ss_entry << "LOG format: [YYYY/MM/DD hh:mm:ss uuu* LEVEL FILE->FUNCTION:LINE] message";
+         ss_entry << "\t\tg3log created log at: " << g3::localtime_formatted(g3::systemtime_now(), "%a %b %d %H:%M:%S %Y") << "\n";
+         ss_entry << "\t\tLOG format: [YYYY/MM/DD hh:mm:ss uuu* LEVEL FILE->FUNCTION:LINE] message";
          ss_entry << "\t\t(uuu*: microseconds fractions of the seconds value)\n\n";
          return ss_entry.str();
       }
--- a/src/filesinkhelper.ipp
+++ b/src/filesinkhelper.ipp
@@ -79,8 +79,8 @@ namespace g3 {
       std::string header() {
          std::ostringstream ss_entry;
          //  Day Month Date Time Year: is written as "%a %b %d %H:%M:%S %Y" and formatted output as : Wed Sep 19 08:28:16 2012
-         ss_entry << "\t\tg3log created log at: " << g3::localtime_formatted(g3::systemtime_now(), "%a %b %d %H:%M:%S %Y") << "\n";
-         ss_entry << "\t\tLOG format: [YYYY/MM/DD hh:mm:ss uuu* LEVEL FILE->FUNCTION:LINE] message";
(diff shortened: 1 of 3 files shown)


src/logmessage.cpp
```diff
@@ -27,35 +27,35 @@ namespace g3 {
    // helper for setting the normal log details in an entry
    std::string LogDetailsToString(const LogMessage& msg) {
       std::string out;
-      out.append("\n" + msg.timestamp() + "\t"
-                 + msg.level() + " [" + msg.file() + "->" + msg.function() + ":" + msg.line() + "]\t");
```
KjellKod: Copying parts of the previous comment from : https://github.com/KjellKod/g3log/pull/143/files

Since this might affect other users I will call on them to voice their opinion. They can diff these and voice any objections:
@jkhoogland , @cstamatopoulos , @ctapmex , @AndreasSchoenle
[original-look.txt](https://github.com/KjellKod/g3log/files/669377/original-look.txt)
[new-look.txt](https://github.com/KjellKod/g3log/files/669378/new-look.txt)

src/filesinkhelper.ipp
```diff
@@ -79,8 +79,8 @@ namespace g3 {
       std::string header() {
          std::ostringstream ss_entry;
          //  Day Month Date Time Year: is written as "%a %b %d %H:%M:%S %Y" and formatted output as : Wed Sep 19 08:28:16 2012
-         ss_entry << "\t\tg3log created log at: " << g3::localtime_formatted(g3::systemtime_now(), "%a %b %d %H:%M:%S %Y") << "\n";
-         ss_entry << "\t\tLOG format: [YYYY/MM/DD hh:mm:ss uuu* LEVEL FILE->FUNCTION:LINE] message";
+         ss_entry << "g3log created log at: " << g3::localtime_formatted(g3::systemtime_now(), "%a %b %d %H:%M:%S %Y") << "\n";
```
KjellKod: Thanks

KjellKod:
👍   pending feedback from the other users. If no-one reply in a day or so I'll go ahead and merge it.

src/logmessage.cpp
cstamatopoulos: I don't see how this will affect many. The fix is only meant to help people that mix g3log with a cout sink and std::cout. Removal of quotation does not bother me at least.

Looking at the txt files, the only visual thing that we lose is the alignment of the messages. In the original look they are mostly aligned which is nice.

rickyzhang82: To be fair, we did not have alignment with/without my change.

If it alighment bother you, we need something like http://stackoverflow.com/questions/2485963/c-alignment-when-printing-cout

cstamatopoulos: I am Ok without it, it was just an observation based on the txt files that were provided.
One seemed aligned one not. And I am only talking about the log comment not the rest info prior to it.

KjellKod: @rickyzhang82 I believe @cstamatopoulos is referring to the alignment in the text log. I.e. the tab alignment  is missing in the PR

screen shot 2016-12-22 at 12 53 26 pm

Could you put it back again please?

src/logmessage.cpp
```diff
@@ -27,35 +27,35 @@ namespace g3 {
...
+      out.append(msg.timestamp() + "\t"
+                 + msg.level() + " [" + msg.file() + "->" + msg.function() + ":" + msg.line() + "]: ");
```
KjellKod: please put back the missing tab character. 

I.e. `]:` --> `}:\t`

src/logmessage.cpp
rickyzhang82: This depends on your text editor how to visualize tab '\t'. In Atom/Vim on my Linux box, neither of them are aligned. Both are ugly.

Personally, I'm not into clutter in the log like the double quote for message and print out how to interprete log message. Most of the time less is more.

KjellKod: As both are ugly I think it makes sense to put back the missing tab character. Once that is done this is good to go. 

Thanks.

rickyzhang82: Gentleman, it is done now. Thanks

rickyzhang82:
I have fixed tab. Why not merge it?

KjellKod:
Geez. How impatient are you? Maybe we aren't on the same timezones? .....

👍

rickyzhang82:
Thanks and Happy holiday!

PR #150, opened 2017-02-13
force /utf-8 for MSVC compiler

0017031:
The source code is already in utf-8; 
but in non-English windows (like mine CP936), it will report warning C4819.
`warning C4819: The file contains a character that cannot be represented in the current code page`

Now in Build.cmake, force utf-8 for MSVC compilers in non-utf8_windows_locale.


KjellKod:
👍  thanks for the contribution

PR #161, opened 2017-03-28
Colorer build and loglevel improvements

KjellKod:
This will replace @ctapmex changes in `colorer`. ref: https://github.com/KjellKod/g3log/pull/
in addition to the code improvement from the PR the following changes were made

- small build structure cleanup
- remove redundant level logic from the old PR
- enhanced function logic to show the client that custom levels *can be added without changing the 
source file*
- added missing unit tests from the colorer's modified functionality changes.


CPackLists.txt
```diff
@@ -6,7 +6,7 @@
 # For more information see g3log/LICENSE or refer refer to http://unlicense.org
 # ============================================================================*/
 
- #  INSTALL( TARGETS g3logger_shared
+ #  INSTALL( TARGETS g3loggerd
```
KjellKod: spelling

src/g3log/atomicbool.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
KjellKod: add header

src/g3log/loglevels.hpp
```diff
@@ -113,13 +144,37 @@ namespace g3 {
...
    // Only safe if done at initialization in a single-thread context
    namespace only_change_at_initialization {
-      // Enable/Disable a log level {DEBUG,INFO,WARNING,FATAL}
+
+      // This sets the logging level status enabled or disabled for a logging level.
+      // If the logging level is NEW it will also ADD the logging level to the set of logging levels.
       void setLogLevel(LEVELS level, bool enabled_status);
+
+      /// Enable log level >= log_level
+      void setLogLevel(LEVELS level);
```
KjellKod: `setLogLevel(LEvELS, enabled_status)` means
1. if `LEVEL` did not exist earlier then add it. 
2. set  the enabled flag for the level

`setLogLevel(LEvELS)` means
1. disable all LEVELS below this one
2. enabled all LEVELS from this and higher. 

Since the functionality is completely different a different name should be used. 
`setHighestLogLevel(LEVEL)` makes sense. 

It MIGHT make sense to have a `addLogLevel(LEVEL)` just to make the API  more readable. it would really just be a func forward to` setLogLevel(LEVEL, true)`

KjellKod: see also https://github.com/KjellKod/g3log/commit/a1748ce66bc1492a4016e916541572434faea03a

PR #168, opened 2017-04-04
logging levels separated between 'only at initialization' and 'log_levels' which is safe during runtime

KjellKod:
The previous namespace for the logging API, `only_change_at_initialization`, is split in two. 

1. 'only at initialization' which contains adding levels and resetting them back to the default. 
2. 'log_levels' where all the calls should be safe to do during runtime.

### DIFF
--- a/src/loglevels.cpp
+++ b/src/loglevels.cpp
@@ -32,7 +32,7 @@ namespace g3 {
 #ifdef G3_DYNAMIC_LOGGING
    namespace only_change_at_initialization {
 
-      void setLogLevel(LEVELS lvl, bool enabled) {
+      void addLogLevel(LEVELS lvl, bool enabled) {
          int value = lvl.value;
          internal::g_log_levels[value] = {lvl, enabled};
       }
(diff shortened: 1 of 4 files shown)


src/loglevels.cpp
```diff
@@ -42,49 +42,76 @@ namespace g3 {
...
 
-
-      std::string printLevels(std::map<int, g3::LoggingLevel> levelsToPrint) {
-
-         std::string levels;
-         for (auto& v : levelsToPrint) {
-            levels += "name: " + v.second.level.text + " level: " + std::to_string(v.first) + " status: " + std::to_string(v.second.status.value()) + "\n";
+      // basically the same as the 'setLevel' but for clarity 
+      // and without any CHECKs for error for illegal level
+      void set(LEVELS level, bool enabled) {
```
KjellKod: enable this on the header API side too

KjellKod: done

src/loglevels.cpp
```diff
@@ -42,49 +42,76 @@ namespace g3 {
...
+      void reset() {
+         g3::internal::g_log_levels = g3::internal::g_log_level_defaults;
+      }
+} // only_change_at_initialization
+
+
+namespace log_levels {
 
-      void setHighestLogLevel(LEVELS enabledFrom) {
+      void setHighest(LEVELS enabledFrom) {
```
KjellKod: The description says
```
      /// Enable log level >= log_level. 
      /// log levels below will be disabled
      /// log levels equal or higher will be enabled.
```
But currently this only disables the one below it. 
*options*
1. change the description
2. push that all above the level should be enabled

KjellKod:
too many issues. not all changes were pushed. closing this until fixed

PR #169, opened 2017-04-05
Check build on windows

ctapmex:
1. fix examples build on windows, if used shared build
2. static library build returned. Not in all cases you can build a shared library


Build.cmake
```diff
@@ -101,35 +101,57 @@ ENDIF()
...
-
-   ADD_LIBRARY(${G3LOG_LIBRARY} SHARED ${SRC_FILES})
-   SET(${G3LOG_LIBRARY}_VERSION_STRING ${VERSION})
-   MESSAGE("Creating ${G3LOG_LIBRARY} VERSION: " ${VERSION})
-   SET_TARGET_PROPERTIES(g3logger PROPERTIES LINKER_LANGUAGE CXX SOVERSION ${VERSION})
-
-   set_target_properties(${G3LOG_LIBRARY} PROPERTIES
+   
+# Create the g3log static library
+   SET(G3LOG_STATIC_LIB g3logger_static)
```
KjellKod: Please add an options flag for building it statically.  Please see Options.cmake. 

Also. Why the ```_static``` suffix?

KjellKod: FYI I am OK with leaving the option to be default ON.

Build.cmake
```diff
@@ -101,35 +101,57 @@ ENDIF()
    ENDIF (MSVC OR MINGW)
 
    set(SRC_FILES ${SRC_FILES} ${SRC_PLATFORM_SPECIFIC})
-
-   # Create the g3log library
    INCLUDE_DIRECTORIES(${LOG_SRC})
-  SET(G3LOG_LIBRARY g3logger)
```
KjellKod: I would keep this so that the logger name is *always*  the same. 

Do you see an issue with building the library with only one of : static or shared?  If so the test library can always just link to ```${G3LOG_LIBRARY}``` it will just be staic *or* shared depending on your cmake configuraiton option (see comment below)

Build.cmake
```diff
@@ -101,35 +101,57 @@ ENDIF()
...
+   ENDIF()
+
+   TARGET_LINK_LIBRARIES(${G3LOG_SHARED_LIB} ${PLATFORM_LINK_LIBRIES})
 
    # Kjell: This is likely not necessary, except for Windows?
-   TARGET_INCLUDE_DIRECTORIES(${G3LOG_LIBRARY} PUBLIC ${LOG_SRC})
+   TARGET_INCLUDE_DIRECTORIES(${G3LOG_SHARED_LIB} PUBLIC ${LOG_SRC})
+   
+   # Set current library for build examples and tests
+   SET(G3LOG_LIBRARY ${G3LOG_SHARED_LIB})
```
KjellKod: ```${G3LOG_LIBRARY}``` should be either the static or the shared.  You can avoid building both by using the option approach.  That is a good way of validating the library since you can choose to build and test them with either one of the two types of library.

KjellKod:
@ctapmex did you see my comments to your pull request?

KjellKod:
Good start. Thanks @ctapmex for pushing g3log to be corrected. 

This PR is  to be replaced by https://github.com/KjellKod/g3log/pull/178

Issue #162, opened 2017-03-29
Adding custom log levels to dynamic logging


Jawan81:
Hi KjellKod
Thanks for your great logger. 
There is one issue I don't find a way to solve: I have defined additional debug levels (e.g. LEVELS ERROR) and want to activate/deactivate them dynamically. But as far I can see one can only do it for the preconfigured log levels (INFO, WARNING etc.) because there seems to be no way to add custom LEVELS to g_log_level_status in loglevels.cpp.
Or am I overlooking something?!
Thanks

Jawan81:
I just noticed you seem to have already done some work on this subject ([commit](https://github.com/KjellKod/g3log/pull/161/commits/26912c4833301f2feccf5c6e0e7c90eb5ed3c241)). When do you think these changes will be merged to master or - even better - when will they become part of a new release?

KjellKod:
With "setLogLevel" you can do exactly that. Please see my recent Pull Request(PR) -  feel free to comment on it. 

https://github.com/KjellKod/g3log/pull/161

The purpose of the PR is to make it easier for coders to understand and use dynamic logging 
I.e enabling/disabling/adding levels of their own choice without having to change the g3log source code.

All the modified and updated APIs have unit tests in the PR that showcase the usage

KjellKod:
As soon as I've reviewed it and hopefully someone else have reviewed it I'll get it merged.

KjellKod:
I'm planning on a release just before merging this since this release changes the scope from c++11 to c++14.

Since the logging level changes are quite important I might turn around and do a small release immediately after this and one more PR is merged

Jawan81:
Thank you for the quick response!

KjellKod:
Until it is merged you can still use the 'setLogLevel' function. If I remember correctly it'll add the level used as argument if it didn't exist already - that part of the API will stay the same

KjellKod:
https://github.com/KjellKod/g3log/pull/161

Jawan81:
The problem with setLogLevel is that it is only save to use it before initialization (as stated by the comment on the only_change_at_initialization namespace: "Only safe if done at initialization in a single-thread context"), but I will need be able to change the log level during runtime.
I solved the problem for now with an extra sink that delegates to the "real" sinks only if the respective log level of the message is active. This extra sink is the only one G3Logs reports to.

KjellKod:
Thanks for the explanation. For the logging level to be synchronized for each message between threads it might need a different design. 

[...]

PR #170, opened 2017-04-06
Levels api changes 2

KjellKod:
The previous namespace for the logging API, only_change_at_initialization, is split in two.

'only at initialization' which contains adding levels and resetting them back to the default.
'log_levels' where all the calls should be safe to do during runtime.

replacing: https://github.com/KjellKod/g3log/pull/168
Addressing. : https://github.com/KjellKod/g3log/issues/162


src/g3log/loglevels.hpp
```diff
@@ -145,44 +145,48 @@ namespace g3 {
...
+
+
+   namespace log_levels {
       /// Enable log level >= log_level. 
       /// log levels below will be disabled
       /// log levels equal or higher will be enabled.
-      /// NOTE: If the level didn't exist prior then it will be added. This is similar to how
-      ///       'setLogLevel' works
-      void setHighestLogLevel(LEVELS level);
+      void setHighest(LEVELS level);
```
Jawan81: Shouldn't the function rather be named `setLowest`? If I call `setLowest(WARNING)` I would expect `WARNING` to be the lowest active level - which it is then -, not the highest like with `setHighest(WARNING).` Or am I misunderstanding something?!

KjellKod: Hah!   Yes. 

Feel like doing a pull request?

Issue #174, opened 2017-04-20
Windows build issues with CMAKE


cstamatopoulos:
It has been sometime since I last pulled but in the past static and shared build used to work correctly.
Now whatever you choose you get a shared build (in windows and probably in any other platform as well).

It is pretty clear from the new cmake that the library is always shared
`ADD_LIBRARY(${G3LOG_LIBRARY} SHARED ${SRC_FILES})`

Instead it should have been something along the lines of 
```
   IF (G3LOG_BUILD_SHARED)
     ADD_LIBRARY(${G3LOG_LIBRARY} SHARED ${SRC_FILES})
   ELSE
     ADD_LIBRARY(${G3LOG_LIBRARY} STATIC ${SRC_FILES})
```
Also it seems that only 1 project is generated and not 2 compared to the past.
Nothing wrong with that but then the flag ADD_BUILD_WIN_SHARED is kind of misleading since you never really add another build but instead the ADD_BUILD_WIN_SHARED flag is only handled to export the symbols. If you want to have a shared build in MSVC that is the only way to export the symbols, unless you go down the path of __declspec(dllexport) and __declspec(dllimport). So if it is MSVC and someone wants to have a shared build this should be set always set
`set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)`
If cmake < 3.4 the cmake should fail since the shared build wont be usable as no symbols are exported.

KjellKod:
Agreed that I should have cleaned it up better.

cstamatopoulos:
This requires changes to 2 files

1st: Options.cmake
```
    option (ADD_BUILD_WIN_SHARED  "Build shared library on Windows" OFF)
       IF(ADD_BUILD_WIN_SHARED)
          MESSAGE("-DADD_BUILD_WIN_SHARED=ON") 
       ELSE() 
          MESSAGE("-DADD_BUILD_WIN_SHARED=OFF") 
      ENDIF()
```
To
```
    option (G3_SHARED_LIB  "Build shared library" ON)
       IF(G3_SHARED_LIB)
          MESSAGE("Building shared library") 
       ELSE() 
          MESSAGE("Building static library") 
      ENDIF()
```

By default we will always build a shared library so this wont affect *Nix, Mac

2nd : Build.cmake

Change 
`   ADD_LIBRARY(${G3LOG_LIBRARY} SHARED ${SRC_FILES})`
To
```
   IF(${G3_SHARED_LIB} STREQUAL "ON")
[...]

Issue #167, opened 2017-04-04
deadlock in aggressive shutdown, re-initialize logger test


KjellKod:
What appears to be a deadlock happens in the 

```
ConceptSink.AggressiveThreadCallsDuringShutdown
```

It's a test that continues to do, from multiple threads,  log calls during a shutdown sequence. Once the shutdown sequence is done the logger is immediately re-initialized. This cycle is done successfully for the whole iteration. It is when the test logging threads are to be `joined` that the deadlock seem to happen.  I.e. it seems to happen in the test threads, not in the logger itself.

KjellKod:
test bug for sure.

KjellKod:
temporarily disabled in pull request: https://github.com/KjellKod/g3log/pull/178

KjellKod:
// From the commented out test ---
// Actually shutting down and re-initializing the logger is not the intention of g3log. 
// the are several initial setups that happen ONCE and the logger relies on the client
// to properly own the logworker that is the key object that receives and direct LOG calls. 
// Making LOG calls thread safe through repeated initialization/shutdowns would come at a 
// high expense of logworker existance synchronization checks. 
/

PR #178, opened 2017-05-02
unit tests running with windows shared dll

AndreasSchoenle:
Changes mainly to CMake files addressing the following issues

- generated definitions stored in the source tree is incompatible with multiple build configurations using the same source checkout. Moved to binary directory
- when using DBUG instead of DEBUG, some compiler errors needed to be fixed
- when compiling g3logger as DLL it should use the runtime DLLs. Otherwise there is problems with allocators (inside the DLL) using a different heap than the deallocators (outside the DLL, i.e. in the test programs). Changed compiler flags to /MD, /MDd to fix this.
- compiler flags where overwritten instead of just adding the /MT, /MTd 
- even with GNU binutils installed, the git version extraction did not work for me due to single quotation marks not being accepted
- needed to set the working directory for the git command as my build directory is outside the source directory.


KjellKod:
There's an issue raised for the DBUG usage. The summary in this pull request mentions that it is fixed but I don't see it

KjellKod:
Overall the pull request seems to target several or all of the recent Windows issues that were raised. 

The summary mentions them but I do not see them addressed in the pull request so I'm assuming it is still in-progress.

Please holler when it is ready for re-review

AndreasSchoenle:
I believe it is all there (now):
- The DBUG usage is in loglevels.cpp and test_unit/test_message.cpp and test_unit/tester_sharedlib.cpp where the DEBUG symbol is used whether or not the DBUG option is on. 
- The /MD vs /MT issue and generated definition is only in CMake
- I added a correction to the generated definitions file (appended it to HEADER_FILES) since your comment so it will be installed with the rest of the headers. Otherwise the change is merely in GenerateMacroDefinitionsFile.cmake writing it to the CMAKE_BINARY_DIR instead of the source tree.
- The git command call to test for the version and build number works for me under windows but the working-directory change for linux out-of-source builds is untested.
Please let me know if anything 
else is still missing? 

Please let me know if there is still something missing.

AndreasSchoenle:
One more comment: Would you mind if I changed all the message( ... ) statements that do not represent warnings in the CMake files to message( STATUS ... )? When running the CMake GUI it actually helps a lot to spot the actual errors when the CMake run fails?

Also: I have added the wchar support again in my repo. It is in the wchar-support branch but I have merged the changes in this pull request (as I could not have tested without them). If it is ok, I will re-create a pull request, when we are done with this one?
Otherwise I can try to base one on master and test without switching to DBUG and with g3logger as static library for now (otherwise all unit tests complain about the memory thing mentioned above) - but I would prefer to do this sequentially.

Also: I have successfully compiled 1.3 with gcc 4.7 on our embedded platform. Only needed to change -std=c++14 back to -std=c++11 and add the other platform specific compiler flags. So if we successfully merge wchar support this time I will definitely switch to 1.3 in production.

AndreasSchoenle:
You mentioned bringing back the static library option. Should I add an option for this? I had removed the shared option because it seemed dead - do you want to keep this option name and bring it back to life? Also: Do I understand correctly that you would want to NOT have both libraries built but either static or DLL?
About the Runtime: IMHO it would still be preferable if the library was linking to the DLL runtime (at least as default) as it seems to be the default for almost all windows DLLs and programs. If g3logger is a static lib and links to the static Runtime it forces all modules using it to do the same to avoid aforementioned heap trouble. 

Except for the dynamic/static thing: Do you still see issues with the PR? Any thoughts on the questions from my previous comment?

AndreasSchoenle:
Added G3_SHARED_LIB option as suggested in #174.

AndreasSchoenle:
OK, seems like I got it together. Appveyor is now testing with DEBUG => DBUG and dynamic logging on (hope this is ok) and runs the unit tests in windows.
It seems like the g3log-FATAL-contract.exe exits with code 3 by design, so I have changed the call such that this does not result in a failed test. 

I have merged all changes to master since then.

Please let me know if anything is missing.

AndreasSchoenle:
I beleive the segfault is not related to the pull request. There seems to be a race condition in pushMessageToLogger() where after the call to isLoggingInitialized() g_logger_instance is used without holding a lock to g_logging_init_mutex(). 

Not sure how to fix without impacting performance. Probably another boolean that is set when testing for initialization which defers destruction?

appveyor.yml
```diff
@@ -28,11 +28,11 @@ before_build:
...
-  - msbuild g3log.sln
+  - cmake -G "Visual Studio 14 2015 Win64" -DADD_G3LOG_UNIT_TEST=ON -DUSE_DYNAMIC_LOGGING_LEVELS=ON -DCHANGE_G3LOG_DEBUG_TO_DBUG=ON ..
+  - msbuild g3log.sln /p:Configuration=Release
   
 # scripts to run after build
 after_build:
-  - cd c:\projects\g3log\build\Release
-  - g3log-FATAL-contract.exe
+  - cmd /c Release\g3log-FATAL-contract.exe || exit /B 0
+  - msbuild RUN_TESTS.vcxproj /p:Configuration=Release
```
KjellKod: Cool thanks.

When this is working later my hope is that appveyor will run all the unit tests. 

Thanks for the DEBUG to DBUG change too.  Great to capture this on this CI

src/loglevels.cpp
```diff
@@ -19,7 +19,11 @@ namespace g3 {
 
 #ifdef G3_DYNAMIC_LOGGING
       const std::map<int, LoggingLevel> g_log_level_defaults = {
-         {g3::kDebugValue, {DEBUG}},
+#if (defined(CHANGE_G3LOG_DEBUG_TO_DBUG))
+		  { DBUG.value, { DBUG } },
+#else
```
KjellKod: It's cramming too much into a single pull request which might mean it's take longer time to get it in but I truly appreciate that you took the time to address this issue

Thanks

test_unit/test_message.cpp
```diff
@@ -189,21 +189,26 @@ TEST(Message, localtime_formatted) {
 #endif // timezone 
 
 
+#if (defined(CHANGE_G3LOG_DEBUG_TO_DBUG))
+#define DEBUG_NAME DBUG
+#else
```
KjellKod: This would be better to keep in the generated options file. Currently you are repeating it in multiple places which should not be needed as long as they are including the generated header file

test_unit/tester_sharedlib.cpp
```diff
@@ -11,6 +11,12 @@
 #include <g3log/logworker.hpp>
 #include "tester_sharedlib.h"
 
+#if (defined(CHANGE_G3LOG_DEBUG_TO_DBUG))
+#define DEBUG_NAME DBUG
+#else
```
KjellKod: Please remove and use the generated header file instead (see comment above)

KjellKod:
The crash is related to this issue. https://github.com/KjellKod/g3log/issues/167

Once the top priority issues are dealt with I'm taking this one on next. Optionally you could add the great disabled test to the test to avoid it since it is already raised as an issue. https://github.com/KjellKod/g3log/issues/167

Build.cmake
```diff
@@ -93,20 +94,32 @@ ENDIF()
    file(GLOB SRC_FILES ${LOG_SRC}/g3log/*.h ${LOG_SRC}/g3log/*.hpp ${LOG_SRC}/*.cpp ${LOG_SRC}/*.ipp)
    file(GLOB HEADER_FILES ${LOG_SRC}/g3log/*.hpp ${LOG_SRC}/*.hpp)
    
+   list( APPEND HEADER_FILES ${GENERATED_G3_DEFINITIONS} )
```
AndreasSchoenle: This is only needed so the file shows up in IDEs (SRC_FILES) and to make sure it is installed later where you use HEADER_FILES to install the includes.

Build.cmake
```diff
@@ -60,8 +61,8 @@ ELSEIF(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
    ENDIF()
 ELSEIF(MSVC)
    set(PLATFORM_LINK_LIBRIES dbghelp)
-   set(CMAKE_CXX_FLAGS_RELEASE "/MT")
-   set(CMAKE_CXX_FLAGS_DEBUG "/MTd")
+   # set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /MD")
```
AndreasSchoenle: Under windows there is no scenario I see where it would be desirable to link against the static Runtime. The DLL runtime is the default, so I removed these flags. As mentioned above using /MTx and building as a DLL or linking against anything that uses the dynamic runtime will result in errors from crtdbg.h in Debug builds (and god knows what in Release builds) as g3logger and the executable or DLL using it will effectively use different heaps.

Options.cmake
```diff
@@ -92,22 +100,13 @@ IF (MSVC OR MINGW)
...
-      MESSAGE("-DDEBUG_BREAK_AT_FATAL_SIGNAL=ON                  Break point for fatal signal is enabled for __DEBUG.") 
-    ELSE() 
-       MESSAGE("-DDEBUG_BREAK_AT_FATAL_SIGNAL=OFF\t\t\tBreak point for fatal signal is disabled") 
-    ENDIF(DEBUG_BREAK_AT_FATAL_SIGNAL)
-
-
-
-# Option for adding windows shared library
-    option (ADD_BUILD_WIN_SHARED  "Build shared library on Windows" OFF)
-       IF(ADD_BUILD_WIN_SHARED)
```
AndreasSchoenle: I have added the G3_SHARED_LIB option as was proposed in another pull request. I can rename it to ADD_BUILD_WIN_SHARED but as we are not adding but it but rather build as either shared or static I thought the name more appropriate

appveyor.yml
```diff
@@ -18,7 +18,7 @@ install:
   # by default, all script lines are interpreted as batch
 
 build:
-  project: g3log\build\g3log.sln
+  project: c:\projects\g3log\build\g3log.sln
```
AndreasSchoenle: In fact the first option did not work for me. I put the absolute path to make it independent of any change of working directory in the script.

src/loglevels.cpp
AndreasSchoenle: OK, should I factor the DEBUG vs DBUG issue out?

test_unit/tester_sharedlib.cpp
AndreasSchoenle: OK, will do.

AndreasSchoenle:
I overlooked #167. Will not look into it for now and disable the test to make sure it does not show as an error any longer.

Options.cmake
```diff
@@ -61,13 +63,21 @@ option (ENABLE_FATAL_SIGNALHANDLING
...
-  MESSAGE("-DENABLE_FATAL_SIGNALHANDLING=ON\tFatal signal handler is enabled")
+   MESSAGE("-DENABLE_FATAL_SIGNALHANDLING=OFF               Fatal signal handler is disabled")
+ELSE() 
+   MESSAGE("-DENABLE_FATAL_SIGNALHANDLING=ON\tFatal signal handler is enabled")
 ENDIF(NOT ENABLE_FATAL_SIGNALHANDLING)
 
+# Option for building as a static or shared library in all platforms
+option (G3_SHARED_LIB  "Build shared library" ON)
+IF(G3_SHARED_LIB)
+   MESSAGE("Building shared library")
```
KjellKod: MESSAGE("-DG3_SHARED_LIB=ON\tBuild shared library")

Options.cmake
```diff
@@ -61,13 +63,21 @@ option (ENABLE_FATAL_SIGNALHANDLING
...
+ELSE() 
+   MESSAGE("-DENABLE_FATAL_SIGNALHANDLING=ON\tFatal signal handler is enabled")
 ENDIF(NOT ENABLE_FATAL_SIGNALHANDLING)
 
+# Option for building as a static or shared library in all platforms
+option (G3_SHARED_LIB  "Build shared library" ON)
+IF(G3_SHARED_LIB)
+   MESSAGE("Building shared library") 
+ELSE() 
+   MESSAGE("Building static library")
```
KjellKod: MESSAGE("-DG3_SHARED_LIB=OFF\tBuild static library")

KjellKod:
@AndreasSchoenle please see my pull request to your pull request. I think that addresses all of the comments that I had (maybe not all the questions). If it looks good please merge with your pull request and then I think this is good to go: https://github.com/AbberiorInstruments/g3log/pull/2

CMakeLists.txt
```diff
@@ -49,10 +49,12 @@
 cmake_minimum_required (VERSION 3.1)
 ENABLE_LANGUAGE(CXX)
 
-if(NOT CMAKE_BUILD_TYPE)
+if(NOT CMAKE_BUILD_TYPE AND NOT (MSVC_IDE OR XCODE))
```
AndreasSchoenle: Those are the multi-configuration generators I know - for those the setting is meaningless and the CMAKE_CONFIGURATION_TYPES variable holds those configurations being generated for the IDE

test_unit/test_io.cpp
```diff
@@ -637,38 +637,38 @@ TEST(CustomLogLevels, AddANonFatal__DidtAddItToEnabledValue) {
 TEST(DynamicLogging, DynamicLogging_IS_ENABLED) {
```
AndreasSchoenle: I merged your pull request. Thanks!

Build.cmake
```diff
@@ -93,20 +94,32 @@ ENDIF()
...
+
+   IF( G3_SHARED_LIB )
+      IF(NOT(CMAKE_VERSION LESS 3.4))
+         set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+      ELSE()
+         message( FATAL "Need CMake version >=3.4 to build shared windows library!" )
+      ENDIF()
+      ADD_LIBRARY(${G3LOG_LIBRARY} SHARED ${SRC_FILES})
+   ELSE()
+      ADD_LIBRARY(${G3LOG_LIBRARY} STATIC ${SRC_FILES})
```
AndreasSchoenle: The list APPEND is IMHO necessary, even on linux. If it did install, maybe it was also left over in the source directory and thus added through the file(GLOB command earlier? 
GLOB will not recursively add files, so even during an in-source build (in a build subdirectory as you do on linux) it should not get added.

Build.cmake
```diff
@@ -120,13 +133,6 @@ ENDIF()
       set_target_properties(${G3LOG_LIBRARY} PROPERTIES MACOSX_RPATH TRUE)
    ENDIF()
 
-   IF(NOT ${ADD_BUILD_WIN_SHARED}.x STREQUAL ".x")
-      IF(NOT(CMAKE_VERSION LESS 3.4) AND MSVC)
```
AndreasSchoenle: With my CMake it does work. I added test messages inside the if statement and it seemed to do what it was supposed to.

Build.cmake
AndreasSchoenle: Creating a static library may well have value. But linking it to the static runtime (which the /MT switches do) will make it unusable in most scenarios: It can then only be linked to an executable or DLL which uses the static runtime as well, which is rare. 
The reason is that it is dangerous at best in MSVC to mix DLL and static runtime or to use several modules (DLLs and Executables) all using the static runtime. Each independent module (DLL or EXE) linking the static runtime introduces its own heap and thus anything allocated dynamically within the module and destroyed outside will lead to heap corruption. 

Therefore all projects I know use the dynamic runtime which is the CMake default.

AndreasSchoenle: Removed the commented out lines.

KjellKod:
@AndreasSchoenle awesome work and great explanations. 

👍

mobileben:
Excellent, thanks for this. This builds the static libraries issue that I had opened! I've successfully built the static libs with this.

src/g3log/loglevels.hpp
```diff
@@ -90,14 +90,10 @@ namespace g3 {
...
 
-#if (defined(CHANGE_G3LOG_DEBUG_TO_DBUG))
-const LEVELS DBUG {g3::kDebugValue, {"DEBUG"}},
-#else
-const LEVELS DEBUG {g3::kDebugValue, {"DEBUG"}},
-#endif
-      INFO {g3::kInfoValue, {"INFO"}},
-      WARNING {g3::kWarningValue, {"WARNING"}},
-      FATAL {g3::kFatalValue, {"FATAL"}};
+const LEVELS G3LOG_DEBUG{g3::kDebugValue, {"DEBUG"}},
```
KjellKod: Ouch. We missed this. It’s broken!

src/g3log/loglevels.hpp
```diff
@@ -90,14 +90,10 @@ namespace g3 {
...
+   INFO {g3::kInfoValue, {"INFO"}},
```
KjellKod: This is now handled on the cpp side : https://github.com/KjellKod/g3log/blob/master/src/loglevels.cpp#L22

PR #181, opened 2017-05-04
powershell use in windows and WORKING_DIRECTORY instead of cd

AndreasSchoenle:
As mentioned in #174, this works for me on windows with out-of-source builds.


KjellKod:
:+1

Issue #186, opened 2017-05-09
Typo in Options.cmake indicating -DG3_SHARED_LIB=ON when it is really OFF


mobileben:
```
IF(G3_SHARED_LIB)
   MESSAGE("-DG3_SHARED_LIB=ON\tBuild shared library") 
ELSE()
   MESSAGE("-DG3_SHARED_LIB=ON\tBuild static library")  
ENDIF()
```

The ```ELSE()``` should have ```OFF``` instead of ```ON```.

KjellKod:
Thanks

KjellKod:
: ref pull request https://github.com/KjellKod/g3log/issues/186

KjellKod:
@mobileben thanks!

PR #187, opened 2017-05-09
Update -DG3_SHARED_LIB=ON to use OFF for the static lib case (#186)

mobileben:
Fixed typo


KjellKod:
Ooops. I forgot you had this PR already. 
Thanks. It was the same content as : https://github.com/KjellKod/g3log/pull/197

👍

PR #189, opened 2017-05-10
Cross-compiling: Version readout fixed

AndreasSchoenle:
Small addition to cmake-changes2: When cross compiling MSVC is of course not set on windows. Also it just checks whether using the MS compiler - will also not be set for mingw. Thus checking for the host OS and determining on which system the command is actually run on seems the way to go.


KjellKod:
:+1: looks good

test_unit/tester_sharedlib.cpp
```diff
@@ -11,12 +11,6 @@
 #include <g3log/logworker.hpp>
 #include "tester_sharedlib.h"
 
-#if (defined(CHANGE_G3LOG_DEBUG_TO_DBUG))
```
KjellKod: Thanks

CMakeLists.txt
```diff
@@ -71,8 +71,8 @@ endif()
 
 
 SET(MAJOR_VERSION 1)
-IF (${VERSION}.x STREQUAL ".x")
-   IF (MSVC)
+IF ( NOT VERSION )
+   IF ( "${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows" )
```
KjellKod: nice change

Build.cmake
```diff
@@ -107,10 +107,12 @@ ENDIF()
    SET(G3LOG_LIBRARY g3logger)
 
    IF( G3_SHARED_LIB )
-      IF(NOT(CMAKE_VERSION LESS 3.4))
-         set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
-      ELSE()
-         message( FATAL "Need CMake version >=3.4 to build shared windows library!" )
+      IF( WIN32 )
```
KjellKod: Why `WIN32` check and not 
```
 IF ( "${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows" )
```?

KjellKod:
👍 

@AndreasSchoenle it looks good. I had only one question that I was curious about. After you answer that I'll go ahead and merge it.

KjellKod:
Tested and needed for cmake-changes4 https://github.com/KjellKod/g3log/pull/190

Issue #194, opened 2017-05-17
undefined reference to g3::log_levels::setHighest(LEVELS)


jmelion:
compiled with GCC 5.4.0 with -DUSE_DYNAMIC_LOGGING=ON

Rest of the library works fine so I do not think this is a linking issue. Is there a pre-processor flag which should be defined but isn't around the section where the implementation of `setHighest` is defined?

KjellKod:
How do you trigger this issue and what is the failure?

KjellKod:
I.e. I have a bunch of test that I am running in `https://github.com/KjellKod/g3log/blob/master/test_unit/test_message.cpp` that is using this function.  

How do I trigger this issue?  (_since when I am building with or without the define it works accordingly to the cmake flag_)

AndreasSchoenle:
@jmelion Most probably the library was compiled without -DUSE_DYNAMIC_LOGGING=ON but the program linking to it was compiled with it. You are on the save side if your program includes the generated header, the compiler will then tell you that the function is not available.

PR #195, opened 2017-05-17
macros now properly set before compiling implementation files

jmelion:
addresses issue #194 

The macros generated by the CMake file would not be seen during the compilation process of the source files.

This would lead to undefined reference linking errors when trying to use methods such as `g3::log_levels::setHighest(LEVELS)` since the appropriate macros such as `G3_DYNAMIC_LOGGING` would not be defined.

This is a quick fix to resolve this issue. It is not very elegant, so may be worth looking for an alternative solution but this solved it for me.

### DIFF
--- a/GenerateMacroDefinitionsFile.cmake
+++ b/GenerateMacroDefinitionsFile.cmake
@@ -30,6 +30,10 @@ FILE(APPEND ${GENERATED_G3_DEFINITIONS} "// CMake induced definitions below. See
 
 FOREACH(definition ${G3_DEFINITIONS} )
    FILE(APPEND ${GENERATED_G3_DEFINITIONS} "#define ${definition}\n")
+   if(definition MATCHES ".* .*")
+   else()
+       add_definitions(-D${definition})
+   endif()
 ENDFOREACH(definition)
 
 MESSAGE("Generated ${GENERATED_G3_DEFINITIONS}")


GenerateMacroDefinitionsFile.cmake
```diff
@@ -30,6 +30,10 @@ FILE(APPEND ${GENERATED_G3_DEFINITIONS} "// CMake induced definitions below. See
 
 FOREACH(definition ${G3_DEFINITIONS} )
    FILE(APPEND ${GENERATED_G3_DEFINITIONS} "#define ${definition}\n")
+   if(definition MATCHES ".* .*")
```
KjellKod: This is causing a huge amount of warnings with `-Wmacro-redefined`

```
/Users/kjell.hedstrom/ws/GitHub/g3log/src/g3log/generated_definitions.hpp:14:9: 
warning: 'G3_DYNAMIC_LOGGING' macro redefined [-Wmacro-redefined]
#define G3_DYNAMIC_LOGGING
        ^
<command line>:1:9: note: previous definition is here
#define G3_DYNAMIC_LOGGING 1
        ^
1 warning generated.
```

Please compile with the examples generated as well as the unit tests.

KjellKod:
Please also explain how the issues are occurring as the unit tests seem to be running fine according to the expectations as decided by the cmake flag

GenerateMacroDefinitionsFile.cmake
AndreasSchoenle: This is because we should EITHER use add_definition (rather, in fact target_compile_definitions( PUBLIC ... )) OR a generated header file.  Like this they will be defined once in the generated header and once on the command line through add_definition()
Assuming everyone uses CMake and properly uses the import script generated by this project, the target_compile_definitions approach has the advantage of not needing a generated file. (using add_definition, this is not the case, the information would be lost as soon as we are not in the context of our cmake project any more). 

However, the HUGE disadvantage would be that as soon as someone does NOT use the cmake import script, there is no way to find out how the library was compiled (except looking in the .cmake file) and even then, the definitions would have to be set by hand to properly use the library. 
The generated header on the other hand is installed alongside the library and just including it makes sure you use the same definitions. I believe it is the better solution.

KjellKod:
@AndreasSchoenle : agreed. The path forward is to use the generated header file.

Since I cannot reproduce the issue (but have my suspicions) I'll let @jmelion explain how to reproduce this before closing the pull request. I need to better understand the circumstances before taking action.

1. Is CMake not used?
2. Compiler, platform, build steps?
3. Is it G3log that fails to compile or is it code linking with G3log?

GenerateMacroDefinitionsFile.cmake
KjellKod: @AndreasSchoenle : agreed. The path forward is to use the generated header file.

Since I cannot reproduce the issue (but have my suspicions) I'll let @jmelion explain how to reproduce this before closing the pull request. I need to better understand the circumstances before taking action.

1. Is CMake not used?
2. Compiler, platform, build steps?
3. Is it G3log that fails to compile or is it code linking with G3log?

AndreasSchoenle: @jmelion Just only saw the original issue. See my comment there. The important thing is that both library and anything linking to it must have the same macros defined. That is what the generated header is for.

KjellKod:
@jmelion make sure you remove any old `g3log/src/g3log/generated_definitions.h` file as it is no longer in that location but if an old file is there it won't be removed from the location automatically

KjellKod:
@jmelion closing this soon unless there is a response. 
Thanks

KjellKod:
Closed due to lack of response

Issue #175, opened 2017-04-21
Wrong time under Linux since v1.3


Jawan81:
We have updated from v1.2 to v1.3 recently, but now the time in the log messages is completely wrong. On my machine the year is displayed to be 2064, on other machines it's 1928.
We use Ubuntu 64bit and 32bit VMs and an embedded target with a custom 32-bit Linux. The error occurs on all machines.
When I debug and step into the timespec_get function in time.cpp (which was reworked from v1.2 to v1.3) on Ubuntu 64-bit here are example values for the local variables:

os_system: 1492790720620630988
os_high_resolution: 1492790720620631654
os: -666
now_ns: 2985581499719334545
ts->tv_sec: 2985581499
ts->tv_nsec: 719334545

This results in the following time in the log output:
2064/08/10 10:11:39 719334

KjellKod:
Hi Jawan81

I've verified the issue and am working on a solution.

KjellKod:
Just FYI: the https://www.epochconverter.com/ gives that the os_system and os_high_resolution are correct. the negative diff will just have an impact on the fraction of seconds and not the year.

KjellKod:
I consider this a high priority issue and hope to have a resolution merged sometime tomorrow. 
I'm sorry for this. This somehow slipped between the verification cracks.

KjellKod:
FYI: I've a solution for this. As soon as I've a pull request up I'll let you know so you can try it.

Jawan81:
Thanks for working on the fix so quickly.

KjellKod:
Family stuff is slowing me down. I've verified a fix for this and it'll be out in a few days. I just can't say when exactly.

Thanks for raising the issue!

KjellKod:
Corrected in: https://github.com/KjellKod/g3log/pull/177

KjellKod:
@Jawan81 Thanks for reporting this.

PR #200, opened 2017-05-18
fixed windows timestamp

AndreasSchoenle:
My old fix for windows high resolution timer. In windows system_time does not have ns precision, only steady_clock (alias for high_resolution_clock) has.
Same principle as before but not relying on high_resolution_clock to be wide enough to store the difference between the clock epochs. Thus the difference is calculated each time when going back to system time.
It might be worthwhile to re-write the formatting functions such that they to the conversion themselves as not to be stuck with the system-time resolution in windows. Not a priority, though, as custom loggers now have access to both.


KjellKod:
Please take a look at this windows CI compiler test
[http://rextester.com/GEZS42474](http://rextester.com/GEZS42474)

The output is something like this
```
date timestamp formatted: 2017/05/18 16:44:24 244657
```

How would that code be formatted on your system?

AndreasSchoenle:
The formatting works the same as for me. Thanks a lot for setting this up. 

The problem is the clock used. Please have a look [at this test.](http://rextester.com/EQSZIP47754)

std::chrono::steady_clock and high_resolution_clock have 100ns resolution resulting in all the times showing up as the same (in fact your formatting only shows microseconds, so in fact, they always would - I am not at work, so I could not look up how to change it to show nanoseconds).
steady_clock and high_resolution_clock have 1ns resolution. As in complex multi-threading multi-processor scenarios a log with ns precision can really help you figure out weird deadlocks and the such, I would really like to have log messages carry such a timestamp - especially as it should not impact performance on the creation side at all.

src/g3log/time.hpp
```diff
@@ -55,6 +56,12 @@ namespace g3 {
    * std::put_time. A possible fix if your c++11 library is not updated is to
    * modify this to use std::strftime instead */
    std::string localtime_formatted(const system_time_point& ts, const std::string& time_format) ;
+
+   inline system_time_point to_system_time(const high_resolution_time_point& ts)
+   {
+	   using namespace std::chrono;
+	   return time_point_cast<system_clock::duration>(system_clock::now() + (ts - high_resolution_clock::now()));
```
KjellKod: The reason why it was changed earlier was that on VMs this creates funky time issues as reported by @Jawan81 in https://github.com/KjellKod/g3log/issues/175

I have not been able to reproduce that issue with this pull request. 
@Jawan81 can you try this pull request in the environments you saw the time issues in to verify that this does not break it?

src/time.cpp
```diff
@@ -141,7 +141,7 @@ namespace g3 {
 
 
    std::string localtime_formatted(const g3::system_time_point& ts, const std::string& time_format) {
-      auto format_buffer = internal::localtime_formatted_fractions(ts, time_format);
+	  auto format_buffer = internal::localtime_formatted_fractions(ts, time_format);
```
KjellKod: formatting is off

src/g3log/logmessage.hpp
```diff
@@ -87,7 +87,7 @@ namespace g3 {
       // Complete access to the raw data in case the helper functions above
       // are not enough.
       //
-      g3::system_time_point _timestamp;
+      g3::high_resolution_time_point _timestamp;
```
KjellKod: It is not needed to change the name. It is hiding implementation details but I think that is intentional.

KjellKod: Maybe changing it to `g3::timestamp` instead.

src/g3log/time.hpp
```diff
@@ -22,7 +22,8 @@
 //          std::string put_time(const struct tm* tmb, const char* c_time_format)
 
 namespace g3 {
-   typedef std::chrono::time_point<std::chrono::system_clock>  system_time_point;
+   typedef std::chrono::time_point<std::chrono::system_clock> system_time_point;
+   typedef std::chrono::time_point<std::chrono::high_resolution_clock> high_resolution_time_point;
```
KjellKod: g3::timepoint

src/g3log/logmessage.hpp
KjellKod: Feel free to come up with a better name.  What I'm striving for is to hide implementation details for retrieving the time point for g3sinks.

src/logmessage.cpp
```diff
@@ -125,7 +125,7 @@ namespace g3 {
 
    LogMessage::LogMessage(const std::string& file, const int line,
                           const std::string& function, const LEVELS& level)
-      : _timestamp(std::chrono::system_clock::now())
+      : _timestamp(std::chrono::high_resolution_clock::now())
```
KjellKod: Maybe `g3::GetTimePoint()`?   I.e. so we stop breaking g3sinks after this change ;)

KjellKod:
@AndreasSchoenle thanks. 
I've added to a fork with system clock and high resolution time so show both with  formatting for:  
ms, us, ns: [http://rextester.com/AXREDO9901](http://rextester.com/AXREDO9901)

KjellKod:
(FYI I removed a bunch of comments)

KjellKod:
@AndreasSchoenle I've updated the code [http://rextester.com/edit/AXREDO9901](http://rextester.com/edit/AXREDO9901). One issue that I see is that the high resolution time stamp sometimes goes backwards. Please see at the bottom

Here's an example output
```
High			System
796896000			796895100
796896000(0)			796895600(500)
796896100(100)			796896100(500)
796896500(400)			796896100(0)
796896500(0)			796896100(0)
796897000(500)			796896600(500)
796896500(-500)			796896600(0)
796897000(500)			796897000(400)
796897000(0)			796897000(0)
796897500(500)			796897000(0)
```

AndreasSchoenle:
Ooops, this is odd - especially since high_resolution_clock is in fact the same as steady_clock in my system. I will look into it tomorrow once more. Thanks a lot for the test code.

AndreasSchoenle:
You are right that the precision is not higher. (At least on this system, will check on some boxes at work tomorrow). The time going back is not the clock, though:
[see here](http://rextester.com/PDYVD11256)
It is in my implementation of to_system_time with the two calls to now(). If one of the clocks advances between them, this will introduce an error of up to 500ns (the step the system clock seems to make). 

Will get back to you tomorrow with my results from work.

AndreasSchoenle:
OK I got it. The system_clock behaves VERY differently on my Win7 system. Maybe it even depends on the hardware? Anyhow, I have modified the test a little:
[http://rextester.com/GDMGCD43687](http://rextester.com/GDMGCD43687)
It now tests clock progress of sys and highres clock (how many for loop cycles before it moves, how much did it move) and it re-implemented the to_system_time function such that the problem mentioned above is no longer there. The output from the compile test linked above shows that both clocks run steady and have a resolution of about 500ns. 
On the other hand, I have attached below the output on my dev system below. It shows that system_clock has a resolution of exactly 1.0001ms (!) while the highres clock has a resolution of about 300ns. I will update the pull request with the to_system_time fix. As Clock resolutions are about half a us, I think the formatting functions using system_clock timepoints with 1us tick duration does not have to be touched. 

It would be even an option to not touch the time stamp of the message and rather calculate it from the high resolution clock (on select systems? But I do not know how to find out how well the clocks perform except with a compile test in CMake). But this would add two additions to the logging side rather than pushing the extra calculation to the sink side.

MY TEST OUTPUT:
```
sytem_clock tick duration:   1e-07 s
highres_clock tick duration: 1e-09 s
g3log test timestamp: %a %b %d %H:%M:%S %Y: Fri May 19 08:45:02 2017

date timestamp formatted: (now) 2017/05/19 08:45:02 377927


Test sysclk clock progress:
1000100 (9532)
2000200 (101872)
3000300 (197128)
4000400 (287562)
5000500 (380379)
6000600 (474921)
7000700 (569299)
8000800 (663840)
9000900 (761496)
10001000 (858048)
11001100 (953105)
12001200 (1044041)
13001300 (1139718)
14001400 (1235431)
15001500 (1331119)
16001600 (1426858)
17001700 (1522611)
18001800 (1618326)
19001900 (1715163)
20002000 (1810860)

Test highres clock progress:
   302 (1)
   603 (4)
   904 (7)
  1205 (10)
  1506 (13)
  1807 (17)
  2109 (20)
  2410 (23)
  2711 (26)
  3012 (30)
  3313 (33)
  3614 (36)
  3916 (39)
  4217 (43)
  4518 (46)
  4819 (49)
  5120 (53)
  5421 (56)
  5723 (59)
  6024 (62)

Test to_system_time:
2017/05/19 08:45:02 381525
2017/05/19 08:45:02 381526
2017/05/19 08:45:02 381526
2017/05/19 08:45:02 381526
2017/05/19 08:45:02 381527
2017/05/19 08:45:02 381527
2017/05/19 08:45:02 381527
2017/05/19 08:45:02 381528
2017/05/19 08:45:02 381528
2017/05/19 08:45:02 381528
2017/05/19 08:45:02 381528
2017/05/19 08:45:02 381529
2017/05/19 08:45:02 381529
2017/05/19 08:45:02 381529
2017/05/19 08:45:02 381530
2017/05/19 08:45:02 381530
2017/05/19 08:45:02 381530
2017/05/19 08:45:02 381531
2017/05/19 08:45:02 381531
2017/05/19 08:45:02 381531
```

src/g3log/time.hpp
```diff
@@ -60,7 +60,10 @@ namespace g3 {
    inline system_time_point to_system_time(const high_resolution_time_point& ts)
    {
 	   using namespace std::chrono;
-	   return time_point_cast<system_clock::duration>(system_clock::now() + (ts - high_resolution_clock::now()));
+	   static auto hrs_now = high_resolution_clock::now();
```
KjellKod: Can you make these function local static members also `const` to further explain there `once and won't change` behavior.

src/g3log/time.hpp
```diff
@@ -60,7 +60,10 @@ namespace g3 {
...
+	   static auto sys_now = system_clock::now();
+
+	   return time_point_cast<system_clock::duration>(sys_now + (ts - hrs_now));
```
KjellKod: Please add a comment explaining the modification also as the reasoning why you time reduction happens is not very obvious.

src/g3log/time.hpp
```diff
@@ -55,6 +56,15 @@ namespace g3 {
    * std::put_time. A possible fix if your c++11 library is not updated is to
    * modify this to use std::strftime instead */
    std::string localtime_formatted(const system_time_point& ts, const std::string& time_format) ;
+
+   inline system_time_point to_system_time(const high_resolution_time_point& ts)
+   {
+	   using namespace std::chrono;
+	   static const auto hrs_now = high_resolution_clock::now();
```
KjellKod: @AndreasSchoenle I've updated to correct the other minor issues that I found with this pull request. 
Could you please here add a comment to explain to the code reader why the time adjustment is done?

Once that is done I think this is ready to be merged. I've tested it on 3 platforms.

AndreasSchoenle: Thanks. I have added a comment, I hope it is a good enough explanation.

KjellKod:
👍 Thanks @AndreasSchoenle this is a good change that should be very useful for Windows users

Issue #203, opened 2017-05-25
Add in support for configurable kMaxMessageSize value


mobileben:
Currently in ```logcapture.cpp``` there is a setting for how much space is available for the buffer available for ```vsnprintf```. 

When exceeded, a ```[...truncated...]``` message is displayed along with the output, which is of course truncated. 

I totally understand the need for constraints and while 2KB works in most cases, for me it's too small. I'm using it to capture server responses so I can use them in test cases without the server. The responses are larger than the limit. 

I'm more than happy to take this as I'm about to pull the latest into my fork and address this since I need it right now. 

The plan would be to add a static method to ```LogCapture``` to set a default value. If there is a worry on any contention of set/get, it could of course be made an ```atomic```. But I have no plans to do that right now.

KjellKod:
I think having the option of changing the value makes sense 

Why not use a CMake option? (Ref the options file)
I would keep it something like this
1. If the option isn't set then keep the size as is
2. If the option is set, make the limit a high value either set by the option or just set to a predefined high value that is sufficient 

With 1 and 2 there is compilation fixation of the value and much easier than adjusting it through a code API

KjellKod:
Thanks @mobileben 

Looking forward to the pull request

mobileben:
The only thing about a compilation version is it makes it a bit more brittle when iterating. Essentially for my MacOS stuff, I build the Release and then ```make install```.  For my iOS stuff, I treat it as a pre-built as well and install it in a pre-built directory I have just for iOS (Release as well). I want to build it once and then "forget about it".

It is more of a PITA to have to rebuild or hold multiple variants of a library, so at least in my case, a code API is more attractive. Especially since the way I use ```g3log``` right now is it is wrapped in a class which takes a configuration file at start to configure it. So I can rapidly modify what it does without any re-building.

Oh, on the side. I did find that wrapping ```g3log``` logging in a class to call ```LogCapture::capturef```.  didn't work out too well. It's probably user error on my behalf. What I wanted to do was "hide" the underlying logger. This includes no header files from ```g3log``` being exposed to the app.

[...]

PR #205, opened 2017-05-29
Add in setMaxMessageSize (#203)

mobileben:
I was going to try and do a unit test, but it only runs on Linux and I'm on a Mac so I couldn't build.

I figured the easiest way to test was to create a sink that understands how to find the ```[...truncated...]``` at the end, which could then be used to verify if we were truncated or not.

I did test this in my current code (both in the client and server). Client is iOS. Server is MacOS.

My tests included:

-Creating a string of 2048 characters using ```std::string(2048, 'A')```. Verify this works
-Created a string of 2049 characters and verify it prints out the ```[...truncated...]```

I then called ```g3::only_change_at_initialization::setMaxMessageSize(200)```

I run the similar tests above (using 200 and 201)

I then called ```g3::only_change_at_initialization::setMaxMessageSize(16000)```

I run the similar tests above (using 16000 and 16001)

It all checked out. Hopefully that is sufficient in testing.

### DIFF
--- a/src/logcapture.cpp
+++ b/src/logcapture.cpp
@@ -19,6 +19,12 @@
 #define SIGNAL_HANDLER_VERIFY() do {} while(0)
 #endif
 
+// MaxMessageSize is message limit used with vsnprintf/vsnprintf_s
+static int MaxMessageSize = 2048;
+
+void g3::only_change_at_initialization::setMaxMessageSize(size_t max_size) {
+   MaxMessageSize = max_size;
+}
 
 
 /** logCapture is a simple struct for capturing log/fatal entries. At destruction the
@@ -60,9 +66,8 @@ LogCapture::LogCapture(const char *file, const int line, const char *function, c
 * See also for the attribute formatting ref:  http://www.codemaestro.com/reviews/18
 */
 void LogCapture::capturef(const char *printf_like_message, ...) {
-   static const int kMaxMessageSize = 2048;
    static const std::string kTruncatedWarningText = "[...truncated...]";
-   char finished_message[kMaxMessageSize];
+   char finished_message[MaxMessageSize];
    va_list arglist;
    va_start(arglist, printf_like_message);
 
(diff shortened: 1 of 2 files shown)


mobileben:
@KjellKod sorry to bother you on this. I don't get the same issue with clang++. So I don't keep committing attempts, is the proper thing to do here to create a ```static const``` and assign it ```MaxMessageSize```?

Example:

`static const int kMaxMessageSize = MaxMessageSize;
`

This will compile fine with ```clang++```

mobileben:
Hmmm, reading about this more, we should re-think about how this gets supported. While clang allowed what I did,  I understand why it shouldn't be done.

Perhaps there should be a compile-time option which allows for the dynamic changing. This way people that don't need this won't incur dynamic memory allocation overhead. 

Ideally I'd love to have a way to be able pick which version gets run dynamically (so I can just use one library), but it's not worth sacrificing usability for others. Unless you can think of a good way to flip between the const array  versus dynamic array version of ```capturef```.

KjellKod:
I can't think of an easy way to do that switch. Compile time option is probably a good option

src/logcapture.cpp
```diff
@@ -60,9 +66,8 @@ LogCapture::LogCapture(const char *file, const int line, const char *function, c
 * See also for the attribute formatting ref:  http://www.codemaestro.com/reviews/18
 */
 void LogCapture::capturef(const char *printf_like_message, ...) {
-   static const int kMaxMessageSize = 2048;
    static const std::string kTruncatedWarningText = "[...truncated...]";
-   char finished_message[kMaxMessageSize];
+   char finished_message[MaxMessageSize];
```
KjellKod: It is illegal c++ to create a stack c-array with a dynamic size. You can get around this by instead using a std::vector as the container and get the c-array from std::vector::data()

mobileben:
I'm cancelling this and submitting a new one. I moved to a new branch using the latest version of g3log.

PR #229, opened 2017-10-26
atomicbool.hpp needs EOF outside of comment to build on VS15

ze897c:
Sorry if I am doing this wrong Kjell, it is my first...

I am trying to use g3log, and everything is great on osx/xcode & linux/gcc, but (what a surprise) windows/VS feels the need to be different

It is simply fixed by adding a newline: I have no idea why VS does not understand the closure of the block, but it does not.
`} // g3'

...added

'// explicit whitespace/EOF for VS15 `


KjellKod:
👍 

Thanks @ze897c

PR #231, opened 2017-11-14
Fix dangling else in LOG and LOGF macros

### LINKED ISSUE #224: Problematic implementation of LOG\LOGF macros
Hi,
There's a basic pitfall in the way LOG\LOGF macros were implemented:
`#define LOG(level) if(g3::logLevel(level)) INTERNAL_LOG_MESSAGE(level).stream()`
If we take the following code:
`if (x)
    LOG(INFO) << "This message";
else
    LOG(INFO) << "That message";`

After the preprocessing stage we get:
`if (x)
    if(g3::logLevel(level)) INTERNAL_LOG_MESSAGE(level).stream() << "This message";
else
    if(g3::logLevel(level)) INTERNAL_LOG_MESSAGE(level).stream() << "That message";   `

As you can see, the compiler has no way to know that the 'else' statement belongs to the first if statement and not the the second one (that comes from the definition of the LOG macro).

The correct way to write this macro would be:
`#define LOG(level) if(!g3::logLevel(level)){\
                                } else INTERNAL_LOG_MESSAGE(level).stream()`
Which will, after preprocessing, turn into:
`if (x)
    if(!g3::logLevel(level)){
    } else INTERNAL_LOG_MESSAGE(level).stream() << "This message";
else
    if(!g3::logLevel(level)){
    } else INTERNAL_LOG_MESSAGE(level).stream() << "That message";`

Which keeps the original logic we're interested in.

maj-tom:
Closes #224

### DIFF
--- a/src/g3log/g3log.hpp
+++ b/src/g3log/g3log.hpp
@@ -145,7 +145,7 @@ namespace g3 {
 
 
 // LOG(level) is the API for the stream log
-#define LOG(level) if(g3::logLevel(level)) INTERNAL_LOG_MESSAGE(level).stream()
+#define LOG(level) if(!g3::logLevel(level)){ } else INTERNAL_LOG_MESSAGE(level).stream()
 
 
 // 'Conditional' stream log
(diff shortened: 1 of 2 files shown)


KjellKod:
Looks good. 

I'm out hunting in the Rockies but will be back in a week and then will go through this again. 

Nice change!

KjellKod:
:+1:  after adding unit test.

PR #234, opened 2017-11-21
Fixing some English/typos and a cyclical header include.

PeasantCodeFarmer:
logcapture.hpp and g3log.hpp included each other.  Fixed by moving include to logcapture.cpp since logcapture.hpp didn't actually use any of it.


KjellKod:
Thanks. Some spelling errors were U.K. English but that change is fine

Nice cleanup!

:+1:

PR #241, opened 2017-12-20
CMake: add install rule

pshurgal:
Now you can use key 'CMAKE_PREFIX_PATH' or 'CMAKE_INSTALL_PREFIX' with cmake and then just
make
make install


KjellKod:
I believed this already worked before. 

What platform were you on before where it did not work?

pshurgal:
I'm using MinGW 5.3.0 and when I try to 'make install' it says that there is not target 'install' at Makefile.
After my changes it works fine.

KjellKod:
Please work with the logic with CPackLists.txt instead in your pull request. 
It seems in the CMakeLists.txt that the corner case of mingw and Windows is not covered. 

```
IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux" OR ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   INCLUDE (${g3log_SOURCE_DIR}/CPackLists.txt)
ENDIF()
```

pshurgal:
@KjellKod I have tried to do as you said, please review. Also I have added mapping of key 'CMAKE_INSTALL_PREFIX' to 'CPACK_INSTALL_PREFIX' for more usability.

CPackLists.txt
```diff
@@ -64,27 +67,33 @@ IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
    message( STATUS "make package" )
 ENDIF()
 
-
-message( STATUS "\nOption to install using 'sudo make install" )
+IF(NOT MINGW)
+   message( STATUS "\nOption to install using 'sudo make install'" )
+ELSE()
```
KjellKod: We can simplify this to just say the message you put in. I.e no `else` or `if` case needed

```
message( STATUS "\nOption to install using 'make install'" )
```

CPackLists.txt
```diff
@@ -64,27 +67,33 @@ IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
...
 #  CPACK_INSTALL_PREFIX
 #  CPACK_PACKAGE_INSTALL_DIRECTORY
+#  CMAKE_INSTALL_PREFIX
 #  CMAKE_INSTALL_HEADERDIR
 #  CMAKE_INSTALL_LIBDIR
 
-
-INCLUDE(CPack)
+IF(NOT MINGW)
+   INCLUDE(CPack)
```
KjellKod: Does CPack not exist on `mingw`? When searching for `mingw cpack CPackLists.txt`. I get several hits from CMake and StackOverflow so I think you should be able to use CPack in your environment

pshurgal: I thought this module is needed for deb package generation.
When I include it under MinGW there is target 'package' in the Make file, but it gives me error about missing NSIS. Now I have installed it and now there is new the error:

CMake Error at D:/g3log/build/cmake_install.cmake:54 (message):
  ABSOLUTE path INSTALL DESTINATION forbidden (by caller):
  D:/g3log/install/include/g3log/active.hpp;D:/g3log/install/include/g3log/atomicbool.hpp;D:/g3log/install/include/g3log/crashhandler.hpp;D:/g3log/install/include/g3log/filesink.hpp;D:/g3log/install/include/g3log/future.hpp;D:/g3log/install/include/g3log/g3log.hpp;D:/g3log/install/include/g3log/logcapture.hpp;D:/g3log/install/include/g3log/loglevels.hpp;D:/g3log/install/include/g3log/logmessage.hpp;D:/g3log/install/include/g3log/logworker.hpp;D:/g3log/install/include/g3log/moveoncopy.hpp;D:/g3log/install/include/g3log/shared_queue.hpp;D:/g3log/install/include/g3log/sink.hpp;D:/g3log/install/include/g3log/sinkhandle.hpp;D:/g3log/install/include/g3log/sinkwrapper.hpp;D:/g3log/install/include/g3log/stacktrace_windows.hpp;D:/g3log/install/include/g3log/std2_make_unique.hpp;D:/g3log/install/include/g3log/stlpatch_future.hpp;D:/g3log/install/include/g3log/time.hpp;D:/g3log/install/include/g3log/g2log.hpp;D:/g3log/install/include/g3log/generated_definitions.hpp

pshurgal:
Ok, what I have found:

https://cmake.org/Wiki/CMake:CPackConfiguration

From this article it follows that:
if you want to set root for package installation you should use 'CPACK_PACKAGING_INSTALL_PREFIX'
if you want to set root where generated packages should be stores you should use 'CPACK_OUTPUT_FILE_PREFIX '

https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html
From that article it follows that:
if you want to set root where artifacts will be installed via 'make install' you should use 'CMAKE_INSTALL_PREFIX'

also absolute paths should not be used at 'INSTALL' section. CMake will append relative paths from 'INSTALL' to 'CPACK_PACKAGING_INSTALL_PREFIX' and 'CMAKE_INSTALL_PREFIX' automatically.

I have tried to rewrite CPackLists.txt according to logic described above. Please review.

pshurgal:
I have tested my solution on WIndows 10 and Ubuntu 17.10 and all worked fine.

pshurgal:
> also absolute paths should not be used at 'INSTALL' section.

This is wrong, on linux you should use absolute path and relative on windows, I'll fix it soon.

KjellKod:
Thanks for the contributiin. 

Due to holidays I will review and test this next week

Merry Christmas

PR #243, opened 2017-12-29
 CMake: CMAKE_INSTALL_PREFIX for Linux

pshurgal:
Now CMAKE_INSTALL_PREFIX and CPACK_PACKAGING_INSTALL_PREFIX both work correctly on Linux, but if both defined during build CMAKE_INSTALL_PREFIX overrides CPACK value.

I also have updated README with MinGW building and installation.


KjellKod:
Nice update :+1:

PR #249, opened 2018-03-02
API suggestion changes

KjellKod:
Since the shared_queue is used outside of g3log. 
It would be useful to improve the api for it for clients that want an upper limit of how much the queue can grow.  
```
   void try_push(T item,  size_t maxSize) {
      {
         std::lock_guard<std::mutex> lock(m_);
         if (size_unsynchronized() < maxSize) {
            queue_.push(std::move(item));
        } else {
           return false;
       }
      }
      data_cond_.notify_one();
      return true;
   }
```


KjellKod:
obsolete with https://github.com/KjellKod/Q

Issue #253, opened 2018-03-16
restoreSignalHandler should restore the original sa_handler


ebertland:
restoreSignalHandler sets sa_handler to SIG_DFL. That's fine for an application that does not register a signal handler, but it if does, then it will be clobbered.

installSignalHandler should store oldact from sigaction (rather than passing nullptr to this argument), and then store that in a map<int, struct sigaction>. Then, restoreSignalHandler gets the old action from the map and passes it to sigaction.

As far as I can tell, there is no way to prevent g3log from clobbering an application or another library's signal handler except to define DISABLE_FATAL_SIGNALHANDLING.

KjellKod:
Good point. 

Please put up a pull request from memory I recognize that it might not be as clear cut. I could remember wrong...

A possible alternative pull request tray would be easy to implement would be to create a “gSignals” modifying function. That way you could remove or add signals to your specific need. Ref: https://github.com/KjellKod/g3log/blob/master/src/crashhandler_unix.cpp

As it is it’s an “all or nothing choice”. Choose g3log’s signal handling or disable it and use your own.

ebertland:
I decided to save and restore the original sigactions. By calling `g3::overrideSetupSignals({});`, I can disable all signal handlers and they are restored to the original ones.

This is working for me, both when I build my stand-alone application, which uses g3log's signal handler, and when I build it as a library and link it to another application that has it's own signal handler and stack trace reporting.

KjellKod:
Addressed in https://github.com/KjellKod/g3log/pull/254

PR #254, opened 2018-03-17
restore original sigaction in restoreSignalHandler

ebertland:
- Save original sigactions in a map called gSavedSigActions
- In restoreSignalHandler, do nothing if there is no saved sigaction.
  If there is a saved sigaction, then re-install it.
- Fixes issue #253

### DIFF
--- a/src/crashhandler_unix.cpp
+++ b/src/crashhandler_unix.cpp
@@ -56,42 +56,6 @@ namespace {
       return (0 == count);
    }
 
-   void reportSigactionError(std::string signal_name) {
-      const std::string error = "sigaction - " + signal_name;
-      perror(error.c_str());
-   }
-
-   void reportSigactionError(int signal_number) {
-      std::string signal_name;
-      const char *signal_name_sz = strsignal(signal_number);
-
-      // From strsignal(3): On some systems (but not on Linux), NULL may instead
-      // be returned for an invalid signal number.
-      if (nullptr == signal_name_sz) {
-         signal_name = "Unknown signal " + std::to_string(signal_number);
-      } else {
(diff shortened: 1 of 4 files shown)


ebertland:
Note that my error handling code for `sigaction() < 0` is untested. I don't know how to cause a sigaction call to fail.

src/crashhandler_unix.cpp
```diff
@@ -58,15 +58,28 @@ namespace {
...
+      if (old_action_it == gSavedSigActions.end()) {
+         // No saved action, so do nothing.
+         return;
+      }
+      
+      if (sigaction(signal_number, &(old_action_it->second), nullptr) < 0) {
+         std::string signal_name;
+         auto signal_name_it = gSignals.find(signal_number);
+         if (signal_name_it == gSignals.end()) {
+            signal_name = signal_name_it->second;
```
KjellKod: Great pull request . This addresses exactly the issue

The “end” handling looks wrong. Once it at the end it can’t access the second and it would crash.

I’ll try it out

KjellKod: Testing that logic is easy, you could just break out the error handling logic into its own function

ebertland: Good catch. I meant `!=`.

ebertland:
I haven't looked at your unit tests. If you can suggest where to add a unit test for reportSigactionError, I will try to do it.

KjellKod:
👍  thanks @ebertland . I had some time and realized that if you haven't developed with g3log and tests it might be hard to get into. 

` cmake -DADD_G3LOG_UNIT_TEST=ON ..` is a start ;)

PR #260, opened 2018-04-22
MinGW / Windows fixes

tkhyn:
Hi and thanks for g3log,

When building on windows (using TDM-GCC 64 compiler), I noticed 2 issues:

1. ``execute_process(COMMAND powershell ...)`` hung => fixed by replacing the call to powershell by a call to cmd
2. the compilation failed because of an undefined reference to ``strerror_s``. Defining ``MINGW_HAS_SECURE_API`` fixed the problem

These 2 changes are included in this PR. I was not able to test with other settings so feel free to let me know if they create any issue.


KjellKod:
:+1:

PR #262, opened 2018-04-27
Fixed ambiguous constructor error.

DigitalInBlue:
* Follow TDD practice. 
* All new and modified functionality should be backed up with unit tests. 
* All new functionality should be backed up with API documentation (API.markdown or README.markdown)


KjellKod:
Long time no see @DigitalInBlue 

Looks good 👍 
Thanks

PR #268, opened 2018-05-15
Export generated defines directory

MaxSavenkov:
This will make the library usable via add_subdirectory from another CMake-based project (without this change, the build fails because it can't find the header with generated defines).


KjellKod:
Thanks. Looks good. Pending verification

KjellKod:
@MaxSavenkov can you describe in this PR how this would be used?  I think you mentioned this earlier but I cannot find it.  I need this information to be able to replicate the issue and the solution to the issue. 

Thanks

MaxSavenkov:
@KjellKod 

Issue: currently, if you try to use g3log as a sub-project in your CMakeLists.txt by using add_subdirectory and target_link_library( my_project PRIVATE g3log ) command, and use g3log in that project by including its header, you will not be able to compile the resulting project, because the main g3log's header would include a generated g3log header, yet the path where generated headers reside is not propagated from sub-project to the main project.

Solution: added a directive for CMake to export the path to generated files (in addition to currently exported path to static header files), so when g3log is imported from another CMakeLists.txt, all necessary include directories will also be imported.

Result: g3log can now be used via add_subdirectory/target_link_library in CMake, without the need to install the library into system or manually specify necessary paths.

PR #281, opened 2018-08-20
Enable relocatable builds

vaivaswatha:
Enable `-fPIC` build of g3log. This is related to https://github.com/Zilliqa/Zilliqa/pull/531


KjellKod:
Good. You probably discovered that you can pass in your own c++ build variables with CMake

Issue #282, opened 2018-08-24
g3log breaks __attribute__ with clang on windows simulating cl


ngladitz:
https://github.com/KjellKod/g3log/blob/14db37ad2313292e4d950fb0c959e92f3182ca48/src/g3log/logcapture.hpp#L53

`__attribute__` is required internally by e.g. clang intrinsic headers which are pulled in by e.g. `<future>` and which are broken by this.

Instead of defining `__attribute__` at all (which seems error prone and effects non-g3log code) perhaps a custom g3log specific macro should be used instead which expands to `__attribute__` where available.

KjellKod:
Interesting. This was introduced 5 years ago and there are many clang compiler users who are using G3log without issues. 

Please give me a simple code example, a main.cpp, which uses G3log and where the functionality is broken by this. 

Please also explain environment, compiler setup and all flags passed in to the compiler
Thanks

ngladitz:
Thank you for looking into this.

Here is a log from a test case that includes the source file, the compiler version (current LLVM windows installer i.e. http://releases.llvm.org/6.0.1/LLVM-6.0.1-win64.exe), invocation and output (shortened to only include two of the errors since the rest look similar).
To reiterate this specifically happens with Clang when it simulates cl (the Microsoft compiler; `__GNUC__` is not defined) not with Clang when it simulates gcc (`__GNUC__` is defined).

```
C:\Users\n.gladitz\Desktop\testcase>type test.cpp 
#include <g3log/logworker.hpp>

C:\Users\n.gladitz\Desktop\testcase>clang-cl --version 
clang version 6.0.1 (tags/RELEASE_601/final)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\LLVM\bin

C:\Users\n.gladitz\Desktop\testcase>clang-cl test.cpp /I. /c 
In file included from test.cpp:1:
In file included from .\g3log/logworker.hpp:16:
In file included from .\g3log/sinkhandle.hpp:11:
In file included from .\g3log/sink.hpp:14:
In file included from .\g3log/future.hpp:28:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.14.26428\include\future:24:
In file included from C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.14.26428\include\ppltasks.h:27:
[...]

PR #283, opened 2018-08-24
Use C++11 generic attribute syntax to avoid issue #282

ngladitz:
This is meant to fix issue #282.

By using C++11 generic attribute syntax I think compiler specific preprocessor conditionals and macro definitions can be entirely avoided in this context.

With my locally available compilers I have confirmed that GCC still produces format violation diagnostics as expected and that Clang simulating cl no longer produces the errors I reported.

Microsoft cl expectedly ignores the unknown attribute.

I am not entirely sure if other compilers ignore the attribute when they don't know it.
It seems to be required since C++17[1] but that may suggest that some compilers may have behaved differently (implementation defined) before.

[1] http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0283r2.html

### DIFF
--- a/src/g3log/logcapture.hpp
+++ b/src/g3log/logcapture.hpp
@@ -47,11 +47,6 @@ struct LogCapture {
 
 
 
-   // Use "-Wall" to generate warnings in case of illegal printf format.
-   //      Ref:  http://www.unixwiz.net/techtips/gnu-c-attributes.html
-#ifndef __GNUC__
-#define  __attribute__(x) // Disable 'attributes' if compiler does not support 'em
-#endif 
 #ifdef _MSC_VER 
 #	if _MSC_VER >= 1400
 #		define G3LOG_FORMAT_STRING _Printf_format_string_
@@ -61,7 +56,10 @@ struct LogCapture {
 #else
 #	define G3LOG_FORMAT_STRING
 #endif
-   void capturef(G3LOG_FORMAT_STRING const char *printf_like_message, ...) __attribute__((format(printf, 2, 3))); // 2,3 ref:  http://www.codemaestro.com/reviews/18
+
+   // Use "-Wall" to generate warnings in case of illegal printf format.
+   //      Ref:  http://www.unixwiz.net/techtips/gnu-c-attributes.html
+   [[gnu::format(printf, 2, 3)]] void capturef(G3LOG_FORMAT_STRING const char *printf_like_message, ...); // 2,3 ref:  http://www.codemaestro.com/reviews/18
 
 
    /// prettifying API for this completely open struct


KjellKod:
@ngladitz thanks for this. Sorry for the late response. It's not forgotten. Matters of a private nature is currently taken precedence over PRs.     

I need to verify this before merging it. I appreciate the hard work to improve g3log

cstamatopoulos:
As mentioned msvc ignores the attribute and generates a warning. If you treat warnings as errors this leads to compilation issues so it would be nice if the attribute was wrapped with some sort of #ifndef _MSC_VER or #if SUPORTED_COMPILERS

KjellKod:
@cstamatopoulus. Shoot! I have no Windows environment to verify this in. If it’s a bother please put up a PR

KjellKod:
@ngladitz see above

ngladitz:
What I mentioned was that MSVC ignored the attribute for me.
In that context I meant "ignored" to mean that there is no warning and that is also the meaning I inferred from the paper I linked.

My warning level however seems to have been insufficiently low as I can now see (at /W3):
```
warning C5030: attribute 'gnu::format' is not recognized
```

Given that there now seem to be two interpretations of what "ignored" actually means here I am not entirely sure if this is a compiler issue or my misinterpretation of the standard.

KjellKod:
I don’t think it really matters. Regardless of whether the compiler issues a warning when it shouldn’t the fastest way to stop issues for Win users is to provide the ifdef protection as suggested earlier

cstamatopoulos:
Since I do not know which compilers support this attribute all I would do is really just move the #endif of line 58 down to 63. Not sure if that is good enough though. I feel that the compilers that actually support this attribute should instead by placed around an #if.

KjellKod:
I could also revert the PR and it can be redone by N

cstamatopoulos:
I feel that moving the #endif is probably as easy and you keep the benefits of the attribute for the other compilers.

ngladitz:
@KjellKod It might indeed not matter for how this is handled but fwiw I've now been told by both Microsoft and GNU developers that "ignored" does not mean that there should be no warnings.

@cstamatopoulos I've found https://en.cppreference.com/w/cpp/feature_test which can be used to check if a compiler supports an attribute but as the title seems to indicate this is not available until C++20 but GCC seems to support it since version 5 [1]. Clang also seems to support it though I am not sure since which version. Both compilers indicate that they support this as an extension in earlier versions of the standard. At the bottom of the page there is an example how one can check for availability of the feature + attribute.

[1] https://www.gnu.org/software/gcc/gcc-5/changes.html

cstamatopoulos:
My personal opinion is that any specific compiler code should be wrapped around #if #endif so that it is not used by other compilers. If in the future we get a way to check for supported attributes that is fine but till then...

I issued a pull request nothing major just moved the #endif as mentioned before. Seems like no one has an issue with how this is handled. My preference does not match my PR but I do not know which specific compilers support this attribute.

Issue #284, opened 2018-08-30
Add synchronous logging mode


ebertland:
I have a use-case for synchronous logging, but I still want to use g3log. In this case, I have multiple libraries that are linked together in one executable. One library uses g3log, while the others do not. I wrote a custom sink to connect g3log's output so that it shows up in the common log file via an API. The order of g3log's messages relative to messages from other libraries in the executable matter, so these messages should be written into the common log synchronously. Also because of the way that the common log API works, I don't think I can send to it from a background thread, at least not easily.

ebertland:
My first thought is to modify Active so that it doesn't spawn a thread. Instead the callback function is executed directly in the send method rather than being pushed into the queue.

```
diff --git a/src/g3log/active.hpp b/src/g3log/active.hpp
index b04ca3c..5bd5833 100644
--- a/src/g3log/active.hpp
+++ b/src/g3log/active.hpp
@@ -45,21 +45,28 @@ namespace kjellkod {
       std::thread thd_;
       bool done_;
 
-
    public:
       virtual ~Active() {
          send([this] { done_ = true;});
-         thd_.join();
+         if (thd_.joinable()) {
+            thd_.join();
+         }
       }
 
       void send(Callback msg_) {
-         mq_.push(msg_);
+         if (thd_.joinable()) {
+            mq_.push(msg_);
+         } else {
+            msg_();
+         }
       }
 
       /// Factory: safe construction of object before thread start
-      static std::unique_ptr<Active> createActive() {
+      static std::unique_ptr<Active> createActive(bool spawn_ = true) {
          std::unique_ptr<Active> aPtr(new Active());
-         aPtr->thd_ = std::thread(&Active::run, aPtr.get());
+         if (spawn_) {
+            aPtr->thd_ = std::thread(&Active::run, aPtr.get());
+         }
          return aPtr;
       }
    };
```

KjellKod:
Hi Jeff,

Before solving the problem I first need to understand the need better. Can
you explain the use case and why synchronous is better than asynchronous?
Please remember that it's still guaranteed FIFO order from the point the
log is added. In essence,. you could potentially crease a synchronous
logger with the same API and have that excangeable for g3log but I am not
sure what the gain would be.

Kjell

ebertland:
Hi Kjell, 

[...]

PR #286, opened 2018-09-01
Add synchronous logging mode (#284)

ebertland:
I know that unit tests and API documentation are needed. I want to see if you like the approach or not, first.

To enable synchronous mode, disable asynchronous mode by calling `g3::only_change_at_initialization::setAsynchronous(false);`
before `g3::LogWorker::createLogWorker();` .

### DIFF
--- a/src/g3log.cpp
+++ b/src/g3log.cpp
@@ -37,6 +37,7 @@ namespace {
    std::once_flag g_initialize_flag;
    g3::LogWorker* g_logger_instance = nullptr; // instantiated and OWNED somewhere else (main)
    std::mutex g_logging_init_mutex;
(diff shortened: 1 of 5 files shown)


KjellKod:
@ebertland thanks for this. Sorry for the late response. It was not forgotten. Matters of a private nature is currently taken precedence over PRs.     

I appreciate the hard work to improve g3log. If the community requests for this feature would grow then I would be interested to bring it it. As it is now I think the changes look good however they are adding complexity that is not (much) requested. 

I would suggest opening up a fork of g3log and I will gladly forward them to that fork if someone else with the same requirement comes along. 

Thank you!

PR #294, opened 2019-01-14
Fix several CMake Issues

BenjaminBeichler:
Since g3log uses some kind of mixed "old" style CMake, it makes it problematic to use in other "modern" cmake projects. Although it is already CMake 3.1 is required, I tries to transform mainly the Build.cmake to conform with modern CMake style, particularly using target specific compile options. This avoids an unwanted pollution of the flags for other parts of bigger projects. 

I tried to keep the whole logic of compile flags for different platforms, moreover g3log now also build cleanly with the musl c-lib. Caused by the lack of other development systems and time, I only checked the CMake scripts in Linux with GCC and Clang but no Windows or OSX.

Moreover, all other cmake-scripts should be reviewed for better compliance.


BenjaminBeichler:
The CI found the first error for Windows: the FindBacktrace Package is not completely MSVC aware. Now I also know, what the dbghelp library is acually linked for. The fix should be easly to insert before the find backtrace the following: 
```
IF(MSVC)
set(backtrace_LIBRARY dbghelp)
ENDIF()
```

BenjaminBeichler:
Sorry for delay, the other pull request just created a merge conflict. I faced several mismatches for option names (which I only fixed for performance test), this should also be revised.

KjellKod:
Not sure about the logic behind moving the header files to a new directory. It doesn’t add any value for the installed files and it only hides files compared to keeping them in the same directory as the source files. 

I know part of the community is in favor of this and many aren’t. Please explain the rationale

BenjaminBeichler:
> Not sure about the logic behind moving the header files to a new directory. It doesn’t add any value for the installed files and it only hides files compared to keeping them in the same directory as the source files.
> 
> I know part of the community is in favor of this and many aren’t. Please explain the rationale

Yes, the rationale was to have a clear separation between headers and sources, but due to the convenience g2log header this separation breaks. CMake also have the recommendation to have only source files in the add_library call, and therefore the separate directory is needed, because the file glob would include a header. Although file glob is also deprecated, I'm not that consequent applying all modern CMake rules. 

However, everything should work without that patch.

The main focus of this Pull Request was to have per target configuration of compiler flags, as it is quite annoying, if you try to compile a c++17 project and several "submodules" are overriding the C++ Standard by there own values like C++11.

If you favor to drop that extra folder thing, I could simply drop that patch.

KjellKod:
@BenjaminBeichler thanks. I'll do my best to get to this PR this week

omcnoe:
This PR looks great and it's made it much easier for me to include g3log with
`add_subdirectory(g3log)
target_link_libraries(my_target PRIVATE g3log)`

I noticed a couple of issues:

First, the version detection with git [here](https://github.com/KjellKod/g3log/blob/1a010de2f62097dede7491ee43549ca1aec7a06b/CMakeLists.txt#L84) fails because it's using CMAKE_SOURCE_DIR instead of CMAKE_CURRENT_SOURCE_DIR.

CMAKE_SOURCE_DIR points at the top level source directory, which obviously fails if g3log is not the top level project.


Second, the build will fail if git is not on the path, because GIT_VERSION will be uninitialized, and so
`math(EXPR VERSION-BASE ${GIT_VERSION}/255)` becomes `math(EXPR VERSION-BASE /255)`. /255 isn't a valid expression for math.

KjellKod:
The first issue can be corrected. The 2nd one is a prerequisite

BenjaminBeichler:
I can fix this, acutally I really missed it, since this is also the "right" modern cmake style :-)

The second problem I actually solved by setting a Version in the parent folder, like this:
`set(VERSION 1.0.0-0)`

or repectively the right code calculated by hand. As you can also keep really one specific version in your project this should not harm anything :-)

KjellKod:
Sweet @BenjaminBeichler 

I'll review this tonight!

garethellis0:
Hi! I actually came here to open this PR. Would really appreciate it getting merged, enforcement of `-stdlib=libstdc++` is causing a bunch of issues for my team (trying to compile this into something that uses `libc++`). Thanks!

KjellKod:
Thanks for the remainder.  Life happened and I lost track of this. I’m putting it on my calendar for tonight

KjellKod:
@BenjaminBeichler this fails to compile on OSX where the current master works fine. 
Steps to reproduce. 
```
cd g3log
// unzip the gtest stuff
mkdir build &&  cd build
cmake ..
make
```
[compile_failure.txt](https://github.com/KjellKod/g3log/files/2872438/compile_failure.txt)

Under what platforms and compilers have you tested this?
I've 
```
OSX
cmake version:  3.12.3
make version: GNU Make 3.81

gcc --version
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
```

KjellKod:
I supect that removal of `-stdlib=libc++` for OSX broke this

michaelbgratton:
I also independently rewrote the cmake for this project so I could use g3 as part of a superbuild like @garethellis0 , but there were too many cases for systems I don't have access to for testing for me to give that cmake to anyone else.

I'd really like to see this change.

KjellKod:
@michaelbgratton 
I totally understand. You can help by reviewing this PR critically 

When I last checked for Cloud/Travis CI for OSX nothing was available. I have some Presedent’s Day time off today so I can take another look

Ideally PR like this should break immediately when a platform support isn’t upheld. Currently Windows and Linux/g++  is under CI test and OSX is done manually

BenjaminBeichler:
Hi,

I haven't tested OSX and I'm also not able to test it, but maybe we can fix this. Why do we need the specific c++ standard library in the first place? Or what is the specific error? It is easier to check for specific features (as e.g. done for the pthread stuff) and do proper handling, insteat of the sledgehammer method to enforce something like the stdlib.

KjellKod:
A std library is required for OSX/Clang or it won’t know how to build library specific features only language specifics.  I.e the price you pay for a configurable compiler is added complexity :/

BenjaminBeichler:
Mhh from what I see, the compile-errors have nothing to do wiith the standard lib, but missing c++11 flags. Could you run make V=1 to show the actually used compile flags ? Somehow the CXX_STANDARD parameter of cmake have bugs in combination with OSX

BenjaminBeichler:
I'm not sure, whether this fixes the problem, but nevertheless this would be needed for proper compiler detection, because of this bug: https://cmake.org/Bug/view.php?id=15355

KjellKod:
@BenjaminBeichler  file from `make VERBOSE=1` is attached
[compile-errors.txt](https://github.com/KjellKod/g3log/files/2876893/compile-errors.txt)

KjellKod:
@BenjaminBeichler compare to master build
[master-build.txt](https://github.com/KjellKod/g3log/files/2876902/master-build.txt)

I haven't spent much time on this but AFAIK the standard library has to be specified for OSX platform. 
You can see the latest PR with my adjustment to it on g3sinks.

KjellKod:
@BenjaminBeichler I think this might have the necessary info for addressing this: https://stackoverflow.com/questions/19774778/when-is-it-necessary-to-use-use-the-flag-stdlib-libstdc

I will take a look at this in detail this PM

BenjaminBeichler:
Ahh, you may retry my lastest push, since the c++ standard requirement was not handed to the test target of your unit test.

BenjaminBeichler:
> @BenjaminBeichler I think this might have the necessary info for addressing this: https://stackoverflow.com/questions/19774778/when-is-it-necessary-to-use-use-the-flag-stdlib-libstdc
> 
> I will take a look at this in detail this PM
But that is the task of CMake and "we" should not be bothered with this :-)

KjellKod:
@BenjaminBeichler agreed in a perfect world .... Glad the cmake bump did the trick though

PR #320, opened 2019-07-27
Ensure LOG* and CHECK* macros are statement-safe

madscientist:
The LOG and LOGF macros had been modified to be statement-safe, but the
LOG_IF, CHECK, LOGF_IF, CHECKF, and CHECK_F macros were all still
unsafe in face of code where single-statement blocks were not enclosed
in `{}`.

For example code like this:
```
  if (!foobar)
      CHECKF(goodness, "badness detected!");
  else
      handle_foobar(foobar);
```
would fail in subtle and possibly dangerous ways.

Fix this by combining multiple if-statements into a single conditional
and inverting the conditions, then adding an empty then-block and moving
the log statement to the else-block.

### DIFF
--- a/src/g3log/g3log.hpp
+++ b/src/g3log/g3log.hpp
@@ -145,19 +145,18 @@ namespace g3 {
 
 
 // LOG(level) is the API for the stream log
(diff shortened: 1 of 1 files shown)


KjellKod:
Looks good. Merge is pending testing.

PR #328, opened 2019-11-20
Set link to local files as relative

jecaro:
The link to the file test_message.cpp was broken. I fixed it with a relative path as it is easier when working on forks. I also changed the link to Options.qmake for the same reason (the initial link wasn't broken thought).


jecaro:
You're welcome !

PR #334, opened 2020-02-25
fix issue with stacktrace_windows not allocating sufficient memory fo…

lederernc:
…r SYMBOL_INFO struct

We encountered an issue in a private code base where printing out the stack after an error would itself cause an access violation in debug builds on Windows. The issue ends up being that the SYMBOL_INFO struct was constructed with insufficient memory to store the symbol name, which far exceeded the allocated amount of memory.

The issue is resolved by allocating more stack memory to allow for the use of up to MAX_SYM_NAME characters in the name field.

### DIFF
--- a/src/stacktrace_windows.cpp
+++ b/src/stacktrace_windows.cpp
@@ -110,7 +110,7 @@ namespace {
 
       DWORD64 displacement64;
       DWORD displacement;
-      char symbol_buffer[sizeof(SYMBOL_INFO) + 256];
+      char symbol_buffer[sizeof(SYMBOL_INFO) + MAX_SYM_NAME];
       SYMBOL_INFO *symbol = reinterpret_cast<SYMBOL_INFO *>(symbol_buffer);
       symbol->SizeOfStruct = sizeof(SYMBOL_INFO);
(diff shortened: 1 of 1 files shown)


src/stacktrace_windows.cpp
```diff
@@ -120,7 +120,7 @@ namespace {
       std::string lineInformation;
       std::string callInformation;
       if (SymFromAddr(GetCurrentProcess(), addr, &displacement64, symbol)) {
-         callInformation.append(" ").append({std::string(symbol->Name), symbol->NameLen});
```
KjellKod: The reason append was used before and not the `std::` constructor was that I was not 100% sure the symbol would not contain null characters. The append function which is almost identical to the constructor for these arguments declare that null characters are fine to add. 

Are you sure that the `symbol` string will never have null characters? Is this documented?
https://en.cppreference.com/w/cpp/string/basic_string/append

KjellKod: Ignore that. I see now that the declaration for `std::string constructor (4)` is identical. 
https://en.cppreference.com/w/cpp/string/basic_string/basic_string

So this change is slightly better as it creates fewer copies (potentially)

Issue #343, opened 2020-05-08
~LogCapture() should be noexcept (false) ?


ripopov:
In a g3log.hpp I read:

> // 'Design By Contract' stream API. For Broken Contracts:
> //         unit testing: it will throw std::runtime_error when a contract breaks
> //         I.R.L : it will exit the application by using fatal signal SIGABRT

So my understanding is that for unit testing I can install setFatalPreLoggingHook that will throw, so I can unit-test contracts with EXPECT_THROW(). However LogCapture() destructor is noexcept (by default), so any attempt to throw from CHECK() will lead to std::terminate.

KjellKod:
Hi, 

g3log uses `SIGABRT` for exiting when a failed contract is discovered. 
For unit testing you can replace the default fatal handling with a custom one using https://github.com/KjellKod/g3log/blob/a6788f89be9b477e156747837909e69fdb009ce6/src/g3log.cpp#L110 using: `setFatalExitHandler(...)`

You can see test_io.cpp which does this frequently using RestoreFileLogger::RestoreFileLogger(...) which sets this.  

https://github.com/KjellKod/g3log/blob/a6788f89be9b477e156747837909e69fdb009ce6/test_unit/testing_helpers.cpp#L117

KjellKod:
The `setFatalPreLoggingHook` was added by community request to trigger specific actions before going into the fatal handling.  It's the `setFatalExitHandler` that you are after for unit testing.

ripopov:
In either case it will be called from destructor,  so 
```
g3::setFatalExitHandler([](g3::FatalMessagePtr){
    throw std::runtime_error("unit test exception");
});
```
will also lead to abort.

So I don't think those comments in source code are correct:

```
// 'Design By Contract' stream API. For Broken Contracts:
//         unit testing: it will throw std::runtime_error when a contract breaks
//         I.R.L : it will exit the application by using fatal signal SIGABRT
#define CHECK(boolean_expression)        \

```

```
// Design By Contract, printf-like API syntax with variadic input parameters.
// Throws std::runtime_eror if contract breaks
#define CHECKF(boolean_expression, printf_like_message, ...)    \
```

```
// Backwards compatible. The same as CHECKF.
// Design By Contract, printf-like API syntax with variadic input parameters.
// Throws std::runtime_eror if contract breaks
#define CHECK_F(boolean_expression, printf_like_message, ...)    \
```

KjellKod:
[...]

PR #349, opened 2020-05-08
Allow throwing from exit handler

ripopov:
For example for unit testing purposes we may want to replace SIGABRT with
throwing an exception, so that unit tests for CHECK()s can be written.

Solves #343

### DIFF
--- a/src/g3log/logcapture.hpp
+++ b/src/g3log/logcapture.hpp
@@ -42,7 +42,7 @@ struct LogCapture {
    // At destruction the message will be forwarded to the g3log worker.
    // In the case of dynamically (at runtime) loaded libraries, the important thing to know is that
    // all strings are copied, so the original are not destroyed at the receiving end, only the copy
-   virtual ~LogCapture();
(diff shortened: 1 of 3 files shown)


KjellKod:
Sweet. Can you add a unit test also that does this?

KjellKod:
Of course. Throwing from a destructor is a big no-no in C++ which I didn’t think of immediately. I’ve to think on this one. 

Regardless, please add unit test for this that has `EXPECT_THROW`

ripopov:
Added unit test.

> Throwing from a destructor is a big no-no in C++ which I didn’t think of immediately. I’ve to think on this one.

In general yes. However,  this usage case, where we utilize destructor of temporary object to do logging, could be a legitimate exception of this rule.

ripopov:
> The way it was approached before is a safer way to do it than mucking with throwing destructors.

Don't agree. If invariant is broken, we should not continue. Like in unit tests I've added, we should not allow code to continue and deference an invalid pointer.

PR #355, opened 2020-06-01
Use CMake to download GoogleTest as part of build

mkilivan:
Having GoogleTest source in project is the hardest approach to keep up
to date and some organizations may not permit this method.

This change is to use CMake to download GoogleTest as part of the
build's configuration step.

It also has the benefit of reducing the number of steps to build the unit
test as unzip is not necessary anymore.


KjellKod:
Looks good. Care to add something similar to g3sinks once this has passed review?

mkilivan:
Sure. Happy to do that.

mkilivan:
GCC 4.x doesn't accept the --std=c++14 switch for C++14 code - it takes --std=c++1y instead. See [here](https://stackoverflow.com/questions/42834844/how-to-get-cmake-to-pass-either-std-c14-c1y-or-c17-c1z-based-on-gcc-vers) for more details.

Using  [CMAKE_CXX_STANDARD](https://cmake.org/cmake/help/v3.2/variable/CMAKE_CXX_STANDARD.html) would be more generic and it should work cross platform.

It seems this flag has been used for the target `${G3LOG_LIBRARY}` in Build.cmake file. You probably want to use a consistent C++ standard for all targets defined in a project, so setting the global the variables would be the most convenient.

```cmake
# Fix behavior of CMAKE_CXX_STANDARD when targeting macOS.
if (POLICY CMP0025)
  cmake_policy(SET CMP0025 NEW)
endif (POLICY CMP0025)

project (g3log CXX)

set (CMAKE_CXX_STANDARD 14)
set (CMAKE_CXX_STANDARD_REQUIRED ON)
set (CMAKE_CXX_EXTENSIONS OFF)
```

Any thoughts?

KjellKod:
Not sure what the CMake policy statement is about, please explain. I prefer fewer details from CMake as CMake is finicky enough. 

The shorter form necessary seems to be 
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

mkilivan:
If you are using CMake and clang to target macOS there is a bug that can cause the CMAKE_CXX_STANDARD feature to not work. But it seems the bug is fixed and this policy is no longer needed if  `cmake_minimum_required(VERSION 3.0)` used. 

Thanks for pointing out. I'll update the patch

KjellKod:
Thanks @mkilivan

PR #356, opened 2020-06-04
Add atomic library for ARM processors

gamelaster:
On my ARM CPU (armv7l, but armel), the compilation hangs on this:
```
[ 60%] Linking CXX shared library libg3logger.so
[ 60%] Built target g3logger
Scanning dependencies of target g3log-FATAL-contract
[ 66%] Building CXX object CMakeFiles/g3log-FATAL-contract.dir/example/main_contract.cpp.o
[ 73%] Linking CXX executable g3log-FATAL-contract
/usr/bin/ld: libg3logger.so.1.3.2-94: undefined reference to `__atomic_fetch_add_8'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/g3log-FATAL-contract.dir/build.make:85: g3log-FATAL-contract] Error 1
make[1]: *** [CMakeFiles/Makefile2:73: CMakeFiles/g3log-FATAL-contract.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```
I find out that the atomic library is required to be linked on ARM processors, and now it compiles fine.
This patch works, although, I'm not much satisfied with ARM detection, sadly, I didn't found any better solution for this. Any improvements are welcomed.

Thanks

EDIT: Also the CI test on macOS failing, kinda not sure how is that possible ☹️

### DIFF
--- a/Build.cmake
+++ b/Build.cmake
@@ -125,6 +125,18 @@ target_compile_options(${G3LOG_LIBRARY} PRIVATE
    $<$<CXX_COMPILER_ID:GNU>:-rdynamic>
 )
 
+# Add atomic library for ARM processors
+IF(UNIX)
+  STRING(FIND ${CMAKE_SYSTEM_PROCESSOR} "arm" arm_res)
+  IF(${arm_res} EQUAL -1)
+    TARGET_LINK_LIBRARIES(${G3LOG_LIBRARY} atomic)
+  ENDIF()
+  STRING(FIND ${CMAKE_SYSTEM_PROCESSOR} "aarch" arm_res)
+  IF(${arm_res} EQUAL -1)
+    TARGET_LINK_LIBRARIES(${G3LOG_LIBRARY} atomic)
+  ENDIF()
+ENDIF()
+
 #cmake -DCMAKE_CXX_COMPILER=clang++ ..
   # WARNING: If Clang for Linux does not work with full c++14 support it might be your
   # installation that is faulty. When I tested Clang on Ubuntu I followed the following


KjellKod:
I’m not sure the detection works as you intended. 
From what I can see when searching for ARM detection on CMake it seems that most people are passing in setup variables when, in fact they are building on ARM. 

Maybe that approach would be a better approach?
Alternatively adding an additional entry in Options.cmake for this. 

——
Before going down this road can you give me the sources that say the atomic library has to be linked? I want to make sure this cannot be resolved by system setup on your end

gamelaster:
> Before going down this road can you give me the sources that say the atomic library has to be linked? I want to make sure this cannot be resolved by system setup on your end

@KjellKod I downloaded the latest version of g3log, and I installed it according the instructions. Sadly, the compilation hangs up on this issue:
```
/usr/bin/ld: libg3logger.so.1.3.2-94: undefined reference to `__atomic_fetch_add_8'
```
After I added into CMake linking of atomic library, it compiled successfully, and the library works without any issues

KjellKod:
@mobileben you more of arm than I do. Can you help out?

@gamelaster  you should also be able to link g3log and the atomic library without doing any changes to the CMake - I mean in case there’s a setup issue or your end

KjellKod:
@AndreasSchoenle and @gimesketvirtadieni you also have ARM experience, what this needed on your end?

KjellKod:
Sorry @gamelaster I don't seem to be getting feedback from @AndreasSchoenle ,  @gimesketvirtadieni  or @mobileben . At this point I am not ready to merge these changes as I do not think it should be resolved within g3log.   If we hear back from these folks then I might reconsider this. 

Please know that you should be able to use CMake commands on the commandline to provide the linking need that you need without changing the repo in this regard.

phryniszak:
Another solution that worked for me is to add atomic library to cmake:

cmake -DCMAKE_SHARED_LINKER_FLAGS='-latomic'

BTW library builds ok with this error. What its failing are examples.
Maybe exclude it from default build?

KjellKod:
You can exclude the examples with CMake options

PR #366, opened 2020-06-24
C++17:  sym link to latest log

xgdgsc:
I find it handy that glog creates symlink to latest log file. This patch requires c++17. So it may not be ok to merge but I think it might help people needing this feature searching in PRs.

### DIFF
--- a/src/filesink.cpp
+++ b/src/filesink.cpp
@@ -10,7 +10,9 @@
 #include "filesinkhelper.ipp"
 #include <cassert>
 #include <chrono>
-
+#if __cplusplus > 201402L
+#include <filesystem>
+#endif
 namespace g3 {
    using namespace internal;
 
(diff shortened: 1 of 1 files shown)


KjellKod:
Lol. It’s early morning here. Of course you have looked at g3sinks ;)
Ha ha ha. So much for good writing when you just wake up

xgdgsc:
I just modified the cmake `set (CMAKE_CXX_STANDARD 17)` and it worked.

KjellKod:
Please don't close this PR. I will get to it but it'll take a bit of time.

KjellKod (changes requested):
@xgdgsc sorry for the late action on this. Can you update the PR now that g3log is c++17 compatible? There are no reasons for the ifdef anymore

src/filesink.cpp
```diff
@@ -10,6 +10,7 @@
 #include "filesinkhelper.ipp"
 #include <cassert>
 #include <chrono>
+#include <filesystem>
```
KjellKod, line 13: dang, apparently c++-17 still has the `experimental/filesystem` on gcc (Linux/OSX). I think that's why it's failing. 

https://stackoverflow.com/questions/45867379/why-does-gcc-not-seem-to-have-the-filesystem-standard-library

lederernc: We use this snippet of code to make getting to std::filesystem work on a variety of compilers/platforms. Not exhaustive, of course.

```suggestion
#if defined(_WIN32)
#    if _MSC_VER >= 1914
#        include <filesystem>
namespace fs = std::filesystem;
#    else
#        include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
#    endif
#elif defined(__GNUC__) && __GNUC__ <= 7
#    include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
#elif defined(__GNUC__) && __GNUC__ >= 8
#    include <filesystem>
namespace fs = std::filesystem;
#elif defined(__clang__) && __clang_major__ <= 8
#    include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
#elif defined(__clang__) && __clang_major__ <= 10
#    include <filesystem>
namespace fs = std::filesystem;
#endif
```

KjellKod: I wonder if this could be used?
Not sure on Windows but I think it works on Linux and OSX
```
// since C++ 20
#include <version>

#ifdef __cpp_lib_filesystem
    #include <filesystem>
    using fs = std::filesystem;
#elif __cpp_lib_experimental_filesystem
    #include <experimental/filesystem>
    using fs = std::experimental::filesystem;
#else
    #error "no filesystem support ='("
#endif
```

xgdgsc: Or use https://github.com/gulrak/filesystem?

lederernc: I think having just made the effort get onto c++17 it makes more sense to start using it.

KjellKod:
Tried it out here: https://github.com/KjellKod/g3log/pull/439 
Windows doesn't seem to have good support for std::filesystem yet

PR #372, opened 2020-07-03
play/tryout:  It is confusing to name the target g3logger. KISS --> g3log

KjellKod:
g3logger was a confusing name. G3log is the name, let's use it.


KjellKod:
Replaced by https://github.com/KjellKod/g3log/tree/g3log.v2.cleanup-install

PR #389, opened 2020-11-17
Alvin/minor fixes and improvements

alvin-777:
- [ ] **TDD** 

New/modified code must be backed down with unit test - preferably TDD style development)

- [ ] **Documentation**

All new/modified functionality should be backed up with API documentation (API.markdown or README.markdown)

**Cross-Platform Testing**
- [x] *Optional:* Local/VM testing: Windows
- [ ] *Optional:* Local/VM testing: OSX
- [x] *Optional:* Local/VM testing: Linux

**Testing Advice**
```bash 
```

**Run Test Alternatives:** 
- Cross-Platform: `ctest`
- Linux: `make test`

### DIFF
--- a/Build.cmake
+++ b/Build.cmake
@@ -8,25 +8,25 @@
 #   and no restrictions or obligations.
 # ===================================================================
 
- # GENERIC STEPS
- SET(LOG_SRC ${g3log_SOURCE_DIR}/src)
+# GENERIC STEPS
+SET(LOG_SRC ${g3log_SOURCE_DIR}/src)
 
- file(GLOB SRC_FILES  ${LOG_SRC}/*.cpp ${LOG_SRC}/*.ipp)
- file(GLOB HEADER_FILES ${LOG_SRC}/g3log/*.hpp)
- 
- list( APPEND HEADER_FILES ${GENERATED_G3_DEFINITIONS} )
- list( APPEND SRC_FILES ${GENERATED_G3_DEFINITIONS} )
+file(GLOB SRC_FILES  ${LOG_SRC}/*.cpp ${LOG_SRC}/*.ipp)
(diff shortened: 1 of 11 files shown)


Build.cmake
```diff
@@ -134,7 +134,7 @@ target_compile_options(${G3LOG_LIBRARY} PRIVATE
 # add GCC specific stuff
 target_compile_options(${G3LOG_LIBRARY} PRIVATE
    # clang/GCC warnings
-   $<$<CXX_COMPILER_ID:GNU>:-rdynamic>
+   $<$<AND:$<CXX_COMPILER_ID:GNU>,$<NOT:$<BOOL:${MINGW}>>>:-rdynamic>
```
alvin-777, line 137: Here is the only line actually modified.

alvin-777:
First commit: fixed a bunch of typos
Second commit: aligned up some indentations and fixed build for MinGW on Windows [(here)](https://github.com/KjellKod/g3log/pull/389/files#r524993218)

Note: even though MinGW builds, it fails the `sink_test` - specifically, it keeps adding messages to the queue, but the sink threads get starved (or maybe stuck?) and can never get any more messages from some point. Eventually, the test runs out of the memory and Windows crashes (the computer I used has 32GB memory, got filled up within 90 seconds).
If I increase the intervals between logs from 1~10 microseconds to 1~10 milliseconds, then it will pass the test immediately. ([this line](https://github.com/KjellKod/g3log/blob/f1eff42b9143af35effe2c78923d2a726119c71c/test_unit/test_sink.cpp#L324))
On Ubuntu, it was no problem either.

KjellKod:
Good find. AFAIK microseconds handling with $MS might be poor. It probably depends on your setup as well.

KjellKod (approved):

PR #390, opened 2020-11-20
Fix compiler warnings and condition check improvements

alvin-777:
- [ ] **TDD** 

New/modified code must be backed down with unit test - preferably TDD style development)

- [ ] **Documentation**

All new/modified functionality should be backed up with API documentation (API.markdown or README.markdown)

**Cross-Platform Testing**
- [x] *Optional:* Local/VM testing: Windows
- [x] *Optional:* Local/VM testing: OSX
- [x] *Optional:* Local/VM testing: Linux

**Testing Advice**
```bash 
```

**Run Test Alternatives:** 
- Cross-Platform: `ctest`
- Linux: `make test`

----

### Fix compiler warnings

- loglevels.cpp - warns about double return if dynamic logging is on
- crashhandler_unix.cpp - warns about unused variable

### Condition check improvements

1. ~Make the condition check more "C++ intuitive"~ (reverted)

For example, people may do something like below:
```cpp
// log the value if pointer is not null
LOG_IF(DEBUG, ptr) << "value: " << ptr->val;
```
This would not work since `ptr != true`, and would possibly confuse them

2. Checking if log level is enabled first.
If it's disabled, there's no need to evaluate the expression (which can be expensive)

### DIFF
--- a/src/crashhandler_unix.cpp
+++ b/src/crashhandler_unix.cpp
@@ -59,7 +59,7 @@ namespace {
    // Dump of stack,. then exit through g3log background worker
    // ALL thanks to this thread at StackOverflow. Pretty much borrowed from:
    // Ref: http://stackoverflow.com/questions/77005/how-to-generate-a-stacktrace-when-my-gcc-c-app-crashes
-   void signalHandler(int signal_number, siginfo_t* info, void* unused_context) {
+   void signalHandler(int signal_number, siginfo_t* /*info*/, void* /*unused_context*/) {
 
       // Only one signal will be allowed past this point
       if (false == shouldDoExit()) {
--- a/src/g3log/g3log.hpp
+++ b/src/g3log/g3log.hpp
@@ -150,12 +150,12 @@ namespace g3 {
 
 // 'Conditional' stream log
 #define LOG_IF(level, boolean_expression)  \
-   if (!g3::logLevel(level) || (boolean_expression) == false) {} else INTERNAL_LOG_MESSAGE(level).stream()
+   if (!g3::logLevel(level) || false == (boolean_expression)) {} else INTERNAL_LOG_MESSAGE(level).stream()
 
 // 'Design By Contract' stream API. Broken Contracts will exit the application by using fatal signal SIGABRT
 //  For unit testing, you can override the fatal handling using setFatalExitHandler(...). See tes_io.cpp for examples
 #define CHECK(boolean_expression)        \
-   if ((boolean_expression) == true) {} else INTERNAL_CONTRACT_MESSAGE(#boolean_expression).stream()
+   if (true == (boolean_expression)) {} else INTERNAL_CONTRACT_MESSAGE(#boolean_expression).stream()
 
 
 /** For details please see this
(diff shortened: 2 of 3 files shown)


src/crashhandler_unix.cpp
```diff
@@ -59,7 +59,7 @@ namespace {
    // Dump of stack,. then exit through g3log background worker
    // ALL thanks to this thread at StackOverflow. Pretty much borrowed from:
    // Ref: http://stackoverflow.com/questions/77005/how-to-generate-a-stacktrace-when-my-gcc-c-app-crashes
-   void signalHandler(int signal_number, siginfo_t* info, void* unused_context) {
+   void signalHandler(int signal_number, siginfo_t* /*info*/, void* /*unused_context*/) {
```
KjellKod, line 62: OK. Unused variables. Accepted

src/g3log/g3log.hpp
```diff
@@ -145,17 +145,17 @@ namespace g3 {
...
 
 // LOG(level) is the API for the stream log
-#define LOG(level) if(!g3::logLevel(level)) {} else INTERNAL_LOG_MESSAGE(level).stream()
+#define LOG(level) if (!g3::logLevel(level)) {} else INTERNAL_LOG_MESSAGE(level).stream()
 
 
 // 'Conditional' stream log
 #define LOG_IF(level, boolean_expression)  \
-   if (false == (boolean_expression) || !g3::logLevel(level)) {} else INTERNAL_LOG_MESSAGE(level).stream()
+   if (!g3::logLevel(level) || !(boolean_expression)) {} else INTERNAL_LOG_MESSAGE(level).stream()
```
KjellKod, line 153: Revert these changes. Although C++ allows a ptr as a valid “Boolean” it’s poor practice. 
I prefer this code to be as is

src/loglevels.cpp
```diff
@@ -132,7 +132,8 @@ namespace g3 {
       int level = log_level.value;
       bool status = internal::g_log_levels[level].status.value();
       return status;
-#endif
+#else
       return true;
```
KjellKod, line 136: Looks good

KjellKod (changes requested):

src/g3log/g3log.hpp
alvin-777: `boolean_expression` checking is reverted. As for checking if log level is enabled before evaluating the expression, I still believe it's necessary.

alvin-777 (commented):
1 is reverted
2 I believe it's necessary

src/g3log/g3log.hpp
```diff
@@ -145,17 +145,17 @@ namespace g3 {
 
 
 // LOG(level) is the API for the stream log
-#define LOG(level) if(!g3::logLevel(level)) {} else INTERNAL_LOG_MESSAGE(level).stream()
+#define LOG(level) if (!g3::logLevel(level)) {} else INTERNAL_LOG_MESSAGE(level).stream()
```
KjellKod, line 148: Format change with my  “space” change not yet removed

src/g3log/g3log.hpp
```diff
@@ -145,17 +145,17 @@ namespace g3 {
...
 
 
 // 'Conditional' stream log
 #define LOG_IF(level, boolean_expression)  \
-   if (false == (boolean_expression) || !g3::logLevel(level)) {} else INTERNAL_LOG_MESSAGE(level).stream()
+   if (!g3::logLevel(level) || (boolean_expression) == false) {} else INTERNAL_LOG_MESSAGE(level).stream()
```
KjellKod, line 153: I’m OK with having the level expression evaluated first. 
Please have the Boolean expression evaluated as it was written 
(false == (Boolean expression)

src/g3log/g3log.hpp
```diff
@@ -145,17 +145,17 @@ namespace g3 {
...
 
 // 'Design By Contract' stream API. Broken Contracts will exit the application by using fatal signal SIGABRT
 //  For unit testing, you can override the fatal handling using setFatalExitHandler(...). See tes_io.cpp for examples
 #define CHECK(boolean_expression)        \
-   if (true == (boolean_expression)) {} else INTERNAL_CONTRACT_MESSAGE(#boolean_expression).stream()
+   if ((boolean_expression) == true) {} else INTERNAL_CONTRACT_MESSAGE(#boolean_expression).stream()
```
KjellKod, line 158: Revert this

src/g3log/g3log.hpp
```diff
@@ -211,17 +211,17 @@ And here is possible output
...
 #define LOGF_IF(level,boolean_expression, printf_like_message, ...) \
-   if (false == (boolean_expression) || !g3::logLevel(level)) {} else INTERNAL_LOG_MESSAGE(level).capturef(printf_like_message, ##__VA_ARGS__)
+   if (!g3::logLevel(level) || (boolean_expression) == false) {} else INTERNAL_LOG_MESSAGE(level).capturef(printf_like_message, ##__VA_ARGS__)
 
 // Design By Contract, printf-like API syntax with variadic input parameters.
 // Calls the signal handler if the contract failed with the default exit for a failed contract. This is typically SIGABRT
 // See g3log, setFatalExitHandler(...) which can be overriden for unit tests (ref test_io.cpp)
 #define CHECKF(boolean_expression, printf_like_message, ...)    \
-   if (true == (boolean_expression)) {} else INTERNAL_CONTRACT_MESSAGE(#boolean_expression).capturef(printf_like_message, ##__VA_ARGS__)
+   if ((boolean_expression) == true) {} else INTERNAL_CONTRACT_MESSAGE(#boolean_expression).capturef(printf_like_message, ##__VA_ARGS__)
```
KjellKod, line 220: All of these needs to be reverted as they were. It’s explicitly put in this order by design

KjellKod (changes requested):

src/g3log/g3log.hpp
alvin-777: This is the only place you didn't put a space after an `if`.
Are you sure that you want to keep it _inconsistent_ with everywhere else?

KjellKod: You are right. For this file. Leave this as is. Thanks for calling it out

KjellKod (approved):

PR #392, opened 2020-12-06
Replace std::result_of with std::invoke_result

Friendly0Fire:
C++17 deprecates `std::result_of<T>` and C++20 completely removes it from the spec in favor of `std::invoke_result<T>`, preventing compilation of g3log if using C++20-level features.

This is the simplest fix which merely replaces all instances of the former with the latter, but it will break compilation on C++14 or older. It would also be possible to make an alias of these templates and use macro compile flags to preserve compatibility with C++14, but I wasn't sure if that would be necessary.

I also took the liberty to use the `*_t` declarations which are much less verbose.


KjellKod:
Thanks. Before I merge this I need to creat a branch and release for “last release with C++14 support”. 

On your end, please update the CI configuration to work with your changes. Once the tests are working I’ll review this PR

Friendly0Fire:
Any idea when this can be merged?

KjellKod:
Thanks for the reminder. I’ve planned to create the release w/tagging and review and this this PR this ongoing week

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
KjellKod, line 1: please remove this file

KjellKod (changes requested):
Kick ass change. Just one tiny file removal and it should be good to go (pending my testing)

.vscode/settings.json
Friendly0Fire: Done!

gir489:
I have confirmed this fixed #337.

PR #399, opened 2021-01-11
Avoid copying LOGLEVELS for g3::logLevel call

lukeocamden:
Don't require copy-constructing a LOGLEVELS instance in order to call g3::logLevel.

This saves an allocation every log statement including when logging is disabled.

### DIFF
--- a/src/g3log/loglevels.hpp
+++ b/src/g3log/loglevels.hpp
@@ -187,6 +187,6 @@ namespace g3 {
 
 #endif
    /// Enabled status for the given logging level
-   bool logLevel(LEVELS level);
+   bool logLevel(const LEVELS& level);
 
 } // g3
--- a/src/loglevels.cpp
+++ b/src/loglevels.cpp
@@ -127,7 +127,7 @@ namespace g3 {
 #endif
 
 
-   bool logLevel(LEVELS log_level) {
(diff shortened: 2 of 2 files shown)


KjellKod:
In the interest of what you are trying to achieve I'd like to call out a couple of things. 

1. https://github.com/KjellKod/g3log/blob/master/src/logmessage.cpp#L155 does a copy. It also needs to do a copy here for memory safety concerns for the community which deals with multiple runtimes. It has to do with handover between runtimes. By copying you ensure that one library runtime does not have to deallocate objects created in another runtime. 
1.  https://github.com/KjellKod/g3log/blob/master/src/logmessage.cpp#L177 is the copy constructor that is invoked here: https://github.com/KjellKod/g3log/blob/master/src/logworker.cpp#L26 the copy is essential to make sure that objects aren't shared across threads. 

In your specific case, depending on your runtime setup etc, it might make sense to mess with "1." for if you every only going to have **one** sink then you don't need to do the copying at 2. either.  Any changes to this however would not be good for the g3log community but of course might make sense to you. 

I can take another look later but AFAIK the changes suggested so far are safe.

KjellKod (approved):
No response to feedback so I'm thinking you are satisfied with these changes

lukeocamden:
Thank you! Much appreciated.

PR #401, opened 2021-03-07
Fix ARM compilation for windows.

englercj:
Added support for ARM in the windows stacktrace utility.

### DIFF
--- a/src/stacktrace_windows.cpp
+++ b/src/stacktrace_windows.cpp
@@ -71,7 +71,17 @@ namespace {
       frame.AddrPC.Mode = AddrModeFlat;
       frame.AddrFrame.Mode = AddrModeFlat;
       frame.AddrStack.Mode = AddrModeFlat;
-#ifdef _M_X64
+#if defined(_M_ARM64)
+      frame.AddrPC.Offset = context->Pc;
+      frame.AddrFrame.Offset = context->Fp;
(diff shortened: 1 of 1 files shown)


KjellKod:
Ideally this would be covered by one of the cloud CIs. I don’t see that as feasible so I’ll approve and merge it. 

Please think about it, if it’s available with ARM on any of the 3 cloud CI we are using them adding ARM support for it would help to future proof ARM support

KjellKod (approved):

KjellKod:
I’m holding off with merging until you you give me the OK that you are confident it’ll work for ARM. 

I’ve no way of verifying this apart from code reviewing

englercj:
Unfortunately I too don't have a machine to verify on, yet. Also unfortunate is the complexity of windows ARM where you generally build on x86 and deploy to ARM.

I've seen similar changes to other stack trace handling code that was updated for ARM though so I'm fairly confident this is correct. I'm also fine leaving it open until we have a chance to verify, but it may be some time for me before I get there. I'm building for ARM64 windows currently, but not ready for testing yet.

englercj:
I think AppVeyor has ARM64 support (from googling) but I'm not familiar enough with that platform you get it running.

KjellKod:
I'm going out on a limb here and making it available. Please consider adding CI support for this later

PR #402, opened 2021-04-04
SOName needs to be less volatile

Shachar:
The SOName of the created library changes with every tiny change to the code. This makes the library impossible to package, as well as a host of other problems. This patch makes the SO name take only the major component of the version.

- [ ] **TDD** 

There is no new code or functionality.

- [ ] **Documentation**

This change applies only to the ABI. There is no change to the API.

**Cross-Platform Testing**
- [ ] *Optional:* Local/VM testing: Windows
- [ ] *Optional:* Local/VM testing: OSX
- [ ] *Optional:* Local/VM testing: Linux

**Testing Advice**
```bash 
```

**Run Test Alternatives:** 
- Cross-Platform: `ctest`
- Linux: `make test`


CMakeLists.txt
```diff
@@ -89,9 +89,11 @@ IF ( NOT VERSION )
    message( STATUS "version base: ${VERSION-BASE}" )
    message( STATUS "version remainder: ${VERSION-REMAINDER}" )
    SET(BUILD_NUMBER ${VERSION-BASE})
-   SET(VERSION ${MAJOR_VERSION}.${MINOR_VERSION}.${BUILD_NUMBER}-${VERSION-REMAINDER})
+   SET(SOFT_VERSION ${MAJOR_VERSION}.${MINOR_VERSION}.${BUILD_NUMBER}-${VERSION-REMAINDER})
```
KjellKod, line 92: Did you mean `soversion`? AFAIK `SOFT_VERSION` is not a reserved keyword in cmake
https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html#prop_tgt:SOVERSION

KjellKod (changes requested):

Shachar:
Revised based on discussion on #403.

KjellKod (approved):

PR #424, opened 2021-12-15
Fix pretty function magic constant overwriting

landwehrj:
This would change the current behavior when printing for gcc/clang.

From:

test_io.cpp->TestBody:639

To:

test_io.cpp->virtual void CustomLogLevels_AddANonFatal_Test::TestBody():639

I replaced \_\_FUNCTION__ with \_\_func__ as the fallback because it seems like that is more standard because it is part of the c99/c++11 standards.

However, maybe \_\_FUNCTION__ should be kept for backwards compatibility sake.

I changed the name of the macro, so it doesn't collide with the compiler versions.

### DIFF
--- a/src/g3log/g3log.hpp
+++ b/src/g3log/g3log.hpp
@@ -30,8 +30,12 @@
 #include <functional>
 
 
-#if !(defined(__PRETTY_FUNCTION__))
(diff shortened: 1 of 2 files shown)


KjellKod (approved):

KjellKod:
Revisiting this and might have to revert some of these changes.

PR #430, opened 2022-01-11
Fixes #428: memory leak in g3::internal::createLogFile

### LINKED ISSUE #428: Memory leak
**Describe the bug**
filesinkhelper.ipp, function g3::internal::createLogFile, in the failure cause, unique_ptr::release is called instead of unique_ptr::reset, which leaks the allocated std::ofstream object - release returns the internal pointer and sets unique_ptr to nullptr WITHOUT deleting the object (i.e. transfers ownership).
See [here](https://en.cppreference.com/w/cpp/memory/unique_ptr/release) for reference.

DlubalSmidJaroslav:
Fixes #428: memory leak in g3::internal::createLogFile


KjellKod (approved):
Thanks for the quick fix and attention to the issue

PR #470, opened 2022-11-30
__SIGFUNC__ no longer default for Windows. __PRETTY__FUNCTION no longer default for gcc/clang

KjellKod:
__SIGFUNC__ no longer default for Windows, It has to be explicitly picked through CMAKE option

### DIFF
--- a/src/g3log/g3log.hpp
+++ b/src/g3log/g3log.hpp
@@ -33,7 +33,7 @@
 #define G3LOG_PRETTY_FUNCTION __FUNCSIG__
 #elif defined(__GNUC__) && defined(PRETTY_FUNCTION) // GCC compatible
 #define G3LOG_PRETTY_FUNCTION __PRETTY_FUNCTION__
-#else // unknown land, let's use basic __FUNCITON__
+#else 
 #define G3LOG_PRETTY_FUNCTION __FUNCTION__
 #endif
 
--- a/src/g3log/g3log.hpp
+++ b/src/g3log/g3log.hpp
@@ -31,13 +31,10 @@
 
 #if defined(_MSC_VER) && (defined(WINDOWS_FUNCSIG)) // Microsoft
 #define G3LOG_PRETTY_FUNCTION __FUNCSIG__
-#pragma message("G3LOG_PRETTY_FUNCTION is set to __FUNCSIG__")
-#elif defined(__GNUC__) // GCC compatible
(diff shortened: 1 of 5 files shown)


KjellKod:
example output with `WINDOWS_FUNCSIG:OFF` ---> in unit test: `Message.UseOverride_4_DetailsWithThreadID_toString`
```
2022/11/30 03:44:57 880438	INFO [test_message.cpp->MyTest::Foo:123]
```

KjellKod:
example output with `PRETTY_FUNCTION:ON` from `examples/g3log-FATAL-sigsegv`
```
INFO [main_sigsegv.cpp->void example_fatal::tryToKillWithAccessingIllegalPointer(std::unique_ptr<std::string>):51]
```
template arguments are expanded.

KjellKod:
Example output with `PRETTY_FUNCTION:OFF` (which is default) from `examples/g2log-FATAL-sigsegv`
```
INFO [main_sigsegv.cpp->tryToKillWithAccessingIllegalPointer:51]
```

KjellKod:
1. Example output with `PRETTY_FUNCTION:OFF` (which is default) from `examples/g2log-FATAL-sigsegv`

```
INFO [main_sigsegv.cpp->tryToKillWithAccessingIllegalPointer:51]
```


2. example output with `PRETTY_FUNCTION:ON` from `examples/g3log-FATAL-sigsegv`
```
// formatted for readability
INFO 
[main_sigsegv.cpp->
void example_fatal::tryToKillWithAccessingIllegalPointer(std::unique_ptr<std::string>):51]
```
template arguments are expanded. 


3. SIGFUNC no longer default for Windows, It has to be explicitly picked through CMAKE option

src/g3log/g3log.hpp
```diff
@@ -29,12 +29,11 @@
 #include <string>
 #include <functional>
 
-
-#if defined(__GNUC__)   // GCC extension compatible
```
KjellKod, line 33: This is easier to read if you look at the plain file rather than the diff

KjellKod:
circleCI runs `PRETTY_FUNCTION:ON` and gitaction runs `PRETTY_FUNCTION:OFF`

KjellKod:
`__PRETTY_FUNCTION__` and `__FUNCSIG__` were  not used before https://github.com/KjellKod/g3log/pull/424

This PR makes sure that the previous default is still `__FUNCTION__` while `__PRETTY_FUNCTION__` and `__FUNCSIG__` can be activated using CMake options.

PR #478, opened 2023-02-20
Fix FileSink dropping messages 

ryanammoury:
# PULL REQUEST DESCRIPTION

Fix FileSink dropping messages introduced by 'optionable buffer to x messages' feature

Commit 6c6122fafc79e92fc94a851b3ff83f87e8b80398 introduced a bug where 99 out of 100 logs were actually dumped

Bug discovered while using two sinks : FileSink and the stdout example sink. I had every messages in stdout sink, but very few in the log file.

# Testing

- [ ] This new/modified code was covered by unit tests. 

- [ ] The CI (Windows, Linux, OSX) are working without issues. 

- [x] Tested manually to check that messages were correctly written to file.

_step 1_

```bash 

```

- Cross-Platform: `ctest`
- Linux: `make test`

### DIFF
--- a/src/filesink.cpp
+++ b/src/filesink.cpp
@@ -67,7 +67,7 @@ namespace g3 {
 
       _write_buffer.append(data);
       if (++_write_counter % _write_to_log_every_x_message == 0) {
-         out << message.get().toString(_log_details_func) << std::flush;
+         out << _write_buffer << std::flush;
          _write_buffer.clear();
       }
    }


src/filesink.cpp
```diff
@@ -67,7 +67,7 @@ namespace g3 {
 
       _write_buffer.append(data);
       if (++_write_counter % _write_to_log_every_x_message == 0) {
-         out << message.get().toString(_log_details_func) << std::flush;
+         out << _write_buffer << std::flush;
```
KjellKod, line 70: damn, this one was ugly. thanks for catching it so quickly.

KjellKod:
Great job, this PR is encapsulated in https://github.com/KjellKod/g3log/pull/479

src/filesink.cpp
ryanammoury: 👍, pleasure contributing to this already great piece of code !

PR #489, opened 2023-05-03
Allow to name g3log worker threads

dimhotepus:
# Allow to name g3log worker threads

Add names to g3log threads to simplify debugging (reasoning about what is this thread created for?)

By default `G3log_Worker#<thread no>`.

# Testing

- [x] This new/modified code was covered by unit tests. 

- [x] The CI (Windows, Linux, OSX) are working without issues. 

_step 1_

```bash 

```

- Cross-Platform: `ctest`
- Linux: `make test`

### DIFF
--- a/src/g3log/active.hpp
+++ b/src/g3log/active.hpp
@@ -27,13 +27,28 @@
 namespace kjellkod {
    typedef std::function<void() > Callback;
 
+#ifdef WIN32
+   typedef std::wstring native_thread_name;
+   const inline native_thread_name DefaultThreadNamePrefix{L"G3log_Worker#"};
+#else
+   typedef std::string native_thread_name;
+   const inline native_thread_name DefaultThreadNamePrefix{"G3log_Worker#"};
+#endif  // !WIN32
+
+   namespace internal {
+      native_thread_name getThreadName(std::thread::native_handle_type handle);
+      void setThreadName(const native_thread_name &threadNamePrefix);
+   }
+
    class Active {
    private:
       Active() : done_(false) {} // Construction ONLY through factory createActive();
       Active(const Active &) = delete;
       Active &operator=(const Active &) = delete;
 
-      void run() {
+      void run(const native_thread_name &threadNamePrefix) {
+         internal::setThreadName(threadNamePrefix);
+
          while (!done_) {
             Callback func;
             mq_.wait_and_pop(func);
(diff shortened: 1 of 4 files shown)


src/g3log/active.hpp
```diff
@@ -57,9 +72,10 @@ namespace kjellkod {
       }
 
       /// Factory: safe construction of object before thread start
-      static std::unique_ptr<Active> createActive() {
+      static std::unique_ptr<Active> createActive(
+         const native_thread_name &threadNamePrefix = DefaultThreadNamePrefix) {
          std::unique_ptr<Active> aPtr(new Active());
-         aPtr->thd_ = std::thread(&Active::run, aPtr.get());
+         aPtr->thd_ = std::thread(&Active::run, aPtr.get(), threadNamePrefix);
```
KjellKod, line 78: I think this concept is interesting but maybe a bit deceiving. While you can change the name, isn't this always just the thread for the active object running in the background? 

Not sure what your purpose is with this. Is it to connect each active object with the sink from a debugging perspectivie? 

Please explain the problem you are trying to solve and how this helps you and others.

KjellKod: I do see the power of this overall but maybe from a different perspective that I need to think about. 

Consider if you could call to set your thread-name for the `LOG` calling thread? 
We already capture the thread_id, and it's optional to include that in the formatting output of the logs. 

I know people have requested an easier way to see which LOG call comes from which thread, and of course a thread_ID, even if it's usable isn't super easy to parse for human eyes compared to plain text. 

I'm not saying you should change the PR towards this. I'm only musing on the power of this functionality and how it could be applied differently. 

Looking forward to your reply about the purpose of the PR

dimhotepus: General idea was to name threads from g3log, as there may be a lot of threads in the app and it is easier to debug them when you have descriptive names on the table.

Anyway, your idea looks good, I'll think about.

KjellKod:
I'll have to come back to this later this week.

KjellKod:
just FYI, I'm still thinking about this one. I'm leaning towards that as it is, we should not merge it. However, I do think it looks promising and I'm grateful for you for sharing this code with us.

KjellKod:
I finally got back to this. 

Since it's thread-naming generally rather than specific for g3log I will close this PR. 
Not sure exactly when something like this would be of use still but it can also be done without messing with UNIX and windows ifdef's 

example that gives the same result. ALTHOUGH. Not sure how this would show up in ones debugger. 
```
#include <string>
#include <mutex>

// Declare the thread_local variable at a scope accessible to both functions
static thread_local std::string my_name;

// Function to set the thread's name
void setThreadName(const std::string& name) {
    static thread_local std::once_flag flag;
    std::call_once(flag, [&]() {
        my_name = name;
    });
}

// Function to get the thread's name
std::string getThreadName() {
    return my_name;
}
```

PR #490, opened 2023-05-03
Fix found by PVS-Studio issues

dimhotepus:
# Some static analyzer fixes

Product page: https://www.viva64.com/en/pvs-studio/
How to use for free: https://www.viva64.com/en/b/0600/

# Testing

- [x] This new/modified code was covered by unit tests. 

- [x] The CI (Windows, Linux, OSX) are working without issues. 

_step 1_

```bash 

```

- Cross-Platform: `ctest`
- Linux: `make test`

### DIFF
--- a/src/filesink.cpp
+++ b/src/filesink.cpp
@@ -92,7 +92,7 @@ namespace g3 {
       ss_change.str("");
 
       std::string old_log = _log_file_with_path;
-      _log_file_with_path = prospect_log;
+      _log_file_with_path = std::move(prospect_log);
       _outptr = std::move(log_stream);
       ss_change << "\n\tNew log file. The previous log file was at: ";
       ss_change << old_log << "\n";
--- a/src/filesinkhelper.ipp
+++ b/src/filesinkhelper.ipp
@@ -51,7 +51,7 @@ namespace g3 {
          return prefix;
       }
 
-      std::string pathSanityFix(std::string path, std::string file_name) {
+      std::string pathSanityFix(std::string path, const std::string &file_name) {
          // Unify the delimeters,. maybe sketchy solution but it seems to work
          // on at least win7 + ubuntu. All bets are off for older windows
          std::replace(path.begin(), path.end(), '\\', '/');
--- a/src/filesinkhelper.ipp
+++ b/src/filesinkhelper.ipp
@@ -88,7 +88,7 @@ namespace g3 {
       std::string createLogFileName(const std::string &verified_prefix, const std::string &logger_id) {
          std::stringstream oss_name;
          oss_name << verified_prefix << ".";
-         if( logger_id != "" ) {
+         if( !logger_id.empty() ) {
             oss_name << logger_id << ".";
          }
          auto now = std::chrono::system_clock::now();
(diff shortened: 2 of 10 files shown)


KjellKod:
At first look this looks like some bold code cleanup. I'll try to review ist next week if my workload goes down. 

Thanks for the look-over and cleanup suggestions

src/filesink.cpp
```diff
@@ -92,7 +92,7 @@ namespace g3 {
       ss_change.str("");
 
       std::string old_log = _log_file_with_path;
-      _log_file_with_path = prospect_log;
```
KjellKod, line 95: nit

src/filesinkhelper.ipp
```diff
@@ -51,7 +51,7 @@ namespace g3 {
          return prefix;
       }
 
-      std::string pathSanityFix(std::string path, std::string file_name) {
```
KjellKod, line 54: nit, but good change. It's within single-threaded scope and only used internally

src/g3log.cpp
```diff
@@ -227,12 +227,13 @@ namespace g3 {
        */
       void pushFatalMessageToLogger(FatalMessagePtr message) {
          if (!isLoggingInitialized()) {
+            const auto& messagePtr = message.get();
```
KjellKod, line 230: Please revert this. Using raw pointers is poor C++ practice and even if it's safe within this limited scope it can easily cause bad code creep. 

There's zero overhead (or completely negligible) with calling a function through unique_ptr compared to calling a function through raw pointer.

src/g3log/logmessage.hpp
```diff
@@ -67,8 +67,8 @@ namespace g3 {
 
       std::string threadID() const;
 
-      void setExpression(const std::string expression) {
-         _expression = expression;
```
KjellKod, line 71: nit

src/logmessage.cpp
```diff
@@ -164,7 +164,7 @@ namespace g3 {
 #endif
       , _file_path(file)
       , _line(line)
-      , _function(function)
```
KjellKod, line 167: nit, most likely a no-op depending on the compiler.

KjellKod (changes requested):
nit == nitpicking, due to personal preference and with zero actual improvement to the code imho. I'm FINE with these `nit` however as I agree with them. It's just not a priority to clean up. Keep as is and thanks for suggesting them.

I only found one area where you need to remove your changes and I put comment on it. It was regarding use of raw pointer instead of unique_ptr

src/g3log.cpp
dimhotepus: Done.

KjellKod: Actually. I misread your code. You are calling from the unique_ptr, just "unboxed" access to it from [MoveOnCopy](https://github.com/KjellKod/g3log/blob/d3f2d11ae603d1c68c3c29d164c3c7b8dbe56f27/src/g3log/moveoncopy.hpp). 

Regardless, you should revert this as 
1. this is only called in the case where the programmer has done a mistake, 
2. or ... the programmer is choosing to not follow setup installation best practices for an unknown reason. 

Regardless, it should be an exceptional case and should not be considered for any possible (?) performance improvement.

dimhotepus: Reverted anyway.

KjellKod:
I’ll re-review this evening

KjellKod:
👍

KjellKod (approved):

Issue #495, opened 2023-05-19
constexpr is not applicable for pointers


CryptoManiac:
Here:

https://github.com/KjellKod/g3log/blob/bad9c58e608498e62ec321f59c78d86aa4ef9e8a/src/g3log/loglevels.hpp#L98

Initialization results with allocation of sub-object on heap. Pointers cannot be used as part of constant expressions.

constexpr should be replaced with const.

CryptoManiac:
```diff
diff --git a/src/g3log/loglevels.hpp b/src/g3log/loglevels.hpp
index 1ddfc63..1d1bb4b 100644
--- a/src/g3log/loglevels.hpp
+++ b/src/g3log/loglevels.hpp
@@ -35,9 +35,6 @@ struct LEVELS {

    LEVELS(const LEVELS& other): value(other.value), text(other.text.c_str()) {}

-   #if __cplusplus >= 202002L
-      constexpr
-   #endif
    LEVELS(int id, const std::string& idtext) : value(id), text(idtext) {}

    bool operator==(const LEVELS& rhs)  const {
@@ -94,11 +91,7 @@ namespace g3 {
    static const int kInternalFatalValue = 2000;
 } // g3

-#if __cplusplus >= 202002L
-   constexpr
-#else
-   const
-#endif
+const
 LEVELS G3LOG_DEBUG{g3::kDebugValue, "DEBUG"},
    INFO {g3::kInfoValue, "INFO"},
    WARNING {g3::kWarningValue, "WARNING"},
@@ -137,11 +130,7 @@ namespace g3 {

 namespace g3 {
    namespace internal {
-      #if __cplusplus >= 202002L
-         constexpr
-      #else
-         const
-      #endif
+      const
       LEVELS CONTRACT {g3::kInternalFatalValue, {"CONTRACT"}},
             FATAL_SIGNAL {g3::kInternalFatalValue + 1, {"FATAL_SIGNAL"}},
             FATAL_EXCEPTION {kInternalFatalValue + 2, {"FATAL_EXCEPTION"}};
```

KjellKod:
@SzGaa please take a look at this.

SzGaa:
Hi,

@CryptoManiac thanks for the report, tomorrow I will verify the issue and check the standard to see if there is an easy way to fix it (other than revert to 'const', but totally agree with @KjellKod this is a syntax sugar, drop it asap if caused problems)

[...]

PR #496, opened 2023-05-19
Update loglevels.hpp

CryptoManiac:
See #495 for details

Issue is not triggered if __cplusplus is less than 202002L on your system.

### DIFF
--- a/src/g3log/loglevels.hpp
+++ b/src/g3log/loglevels.hpp
@@ -35,9 +35,6 @@ struct LEVELS {
 
    LEVELS(const LEVELS& other): value(other.value), text(other.text.c_str()) {}
 
-   #if __cplusplus >= 202002L
-      constexpr
-   #endif
    LEVELS(int id, const std::string& idtext) : value(id), text(idtext) {}
 
    bool operator==(const LEVELS& rhs)  const {
@@ -94,11 +91,7 @@ namespace g3 {
    static const int kInternalFatalValue = 2000;
 } // g3
 
-#if __cplusplus >= 202002L
-   constexpr
-#else
-   const
-#endif
+const
 LEVELS G3LOG_DEBUG{g3::kDebugValue, "DEBUG"},
    INFO {g3::kInfoValue, "INFO"},
    WARNING {g3::kWarningValue, "WARNING"},
@@ -137,11 +130,7 @@ namespace g3 {
 
 namespace g3 {
    namespace internal {
-      #if __cplusplus >= 202002L
-         constexpr
-      #else
-         const
-      #endif
+      const
       LEVELS CONTRACT {g3::kInternalFatalValue, {"CONTRACT"}},
             FATAL_SIGNAL {g3::kInternalFatalValue + 1, {"FATAL_SIGNAL"}},
             FATAL_EXCEPTION {kInternalFatalValue + 2, {"FATAL_EXCEPTION"}};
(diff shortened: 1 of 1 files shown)


KjellKod:
@SzGaa  please review this.

KjellKod:
I think this is fine. Running tests now. `constexpr` vs const isn't really a huge win regardless outside of making it more up-to-date. I'm planning to merge this once it's all gone through CI

KjellKod:
Thank you @CryptoManiac for this insight and update

src/g3log/loglevels.hpp
```diff
@@ -94,11 +91,7 @@ namespace g3 {
    static const int kInternalFatalValue = 2000;
 } // g3
 
-#if __cplusplus >= 202002L
-   constexpr
-#else
-   const
-#endif
+const
```
KjellKod, line 94: please make this the same line as `LEVELS G3LOG_DEBUG{g3::kDebugValue, "DEBUG"},`

Ref how it was earlier https://github.com/KjellKod/g3log/commit/cc0fb7c1ea7d7cc2bc5d8766b18b190ba74b70f9

KjellKod (changes requested):

CryptoManiac:
> I think this is fine. Running tests now. `constexpr` vs const isn't really a huge win regardless outside of making it more up-to-date. I'm planning to merge this once it's all gone through CI

This testing doesn't have a lot of sense because `const` vs `constexpr` have fundamental differences which are nothing to do with either performance or being up to date. The `const` identifiers are variables. Read-only, yes, but still allocated as usual. The same can't be said about constexpr expressions, they only exist at compile-time and there is no such thing in the generated code. Therefore, there is a lot of limitations on what you can or can't do with them. For `constexpr` identifier any input data must be known at compile time, meaning that there is no way to allocate dynamic memory for them or whatever. They're good for switches and math equations or other things like that. You can use the result of constexpr expression as a template argument, for example, while the same can't be done with `const` identifiers.

SzGaa:
> The `const` identifiers are variables. Read-only, yes, but still allocated as usual. The same can't be said about constexpr expressions, they only exist at compile-time and there is no such thing in the generated code.

Compiler is not forced to use constexpr in compile context only (also not guaranteed that it will use that way even if could be possible), just allows it to be used in places where only compile time constants would be applicable.
Also constexpr implies constness.

[constexpr specifier](https://en.cppreference.com/w/cpp/language/constexpr)

CryptoManiac:
> Compiler is not forced to use constexpr in compile-time context only (also not guaranteed that it will use that way even if could be possible), just allows it to be used in places where only compile time constants would be applicable. 

It allows them to be used in places like that because it does provide a guarantee that the expression is deterministic and the required data is known at compile-time. Therefore it _may be_ evaluated in compile-time. Of course, for the sake of common sense, compiler won't actually evaluate expressions unless they're something very simple. But this has no visible side effects since the said limitations still stand. No heap allocations, no exceptions and only literal value types. It's also kind of understandable why `goto` is not supported, since though they're quite possible to implement, it would be really hard to actually do such a thing. 

> Also constexpr implies constness.

Well, Shakespeare was born in England, he wrote in english and spoke in english as well. Therefore, Shakespeare was english. Of course they are const since they're deterministic and their input data is const as well. :)

KjellKod (approved):

PR #506, opened 2023-11-27
trying out codespaces

KjellKod:
ref: https://github.com/microsoft/vscode-remote-try-cpp/tree/main


KjellKod (commented):
works like a charm.

PR #513, opened 2023-12-06
Mostly a refactor of tests

KjellKod:
# PULL REQUEST DESCRIPTION

1. added refactor of unit tests where fatal handling was separated out. 
2. FATAL signal handling is when the process is in a dire situation. no malloc or similar is ever within the `async-signal-safe` space and yet we must push, or try to push what we have to the logging sink. Mostly it'll work and if it didn't ... heck it was worth a shot.  ---> At least a small cleanup is to avoid perror in the signal handler.


src/crashhandler_windows.cpp
```diff
@@ -171,6 +171,12 @@ namespace g3 {
          raise(signal_number);
       }
 
+      // FYI: Concept of async-signal-safe operations does not exist on windows
+      // we stick to perror for lack of better alternatives.
+      size_t writeErrorMessage(const char* message) {
+         perror(message)
```
KjellKod, line 177: semicolon missing -- it sucks that there's no easy way to spin up vscontainer with windows afaik

PR #518, opened 2024-02-05
Update docs.yml for EOL github action versions

KjellKod:


.github/workflows/docs.yml
```diff
@@ -13,10 +13,10 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
```
KjellKod, line 17: V5?

PR #537, opened 2024-08-28
Update ctest.yml

KjellKod:
updated for ctest issues on osx


KjellKod:
yml files are not triggering matrix build actions.

